PR #3, opened 2024-07-15
Add internationalization

david-mears-2:
Add internationalization.

The proposed value of this is that, if we want to add another language in the future, it's best to have the internationalization in from the start.

There are some side-benefits like e.g. if you want to update the name of a certain page, or of the whole app, you can do that just in one place and then anywhere that mentions the name will be in sync.

A cost is that you no longer see the actual text you meant in your template. So we should make all i18n keys pretty transparent.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -26,5 +26,5 @@ jobs:
         uses: codecov/codecov-action@v4
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
-          slug: j-idea/daedalus-web-app
+          slug: jameel-institute/daedalus-web-app
           fail_ci_if_error: true


david-mears-2:
Todo: https://nuxt.com/docs/guide/concepts/auto-imports#auto-import-from-third-party-packages

tests/vitestSetup.ts
```diff
@@ -0,0 +1,13 @@
...
+
+// https://github.com/nuxt-modules/i18n/issues/2637
+// https://github.com/nuxt/test-utils/issues/566
+// https://github.com/nuxt/test-utils/issues/585
+const i18n = createI18n({
+  legacy: false,
+  locale: "en",
+  messages: { en },
+});
+config.global.plugins.push(i18n);
```
EmmaLRussell, line 13: Presumably if we ever get a second language we can have tests for that which override this global plugin config?

EmmaLRussell (commented):
LGTM!

I think it could be made somewhat more typesafe (so you'd get a transpilation error with a typo in a translation key name): https://www.i18next.com/overview/typescript
..though this isn't with the nuxt specific package.. 
Not sure how far this will get you: https://i18n.nuxtjs.org/docs/v7#typescript

I guess we'll also eventually have to figure out how translations happen for stuff coming from the R API e.g. column names and errors. Do those translations originate in the API, so the front end can be totally ignorant of those? If so we will need a way either for the front end to tell the API what the current language is (that's how we did it in HINT) or for the API to communicate all relevant translation values to the front end.

PR #9, opened 2024-07-29
Jidea 38 containerisation wip

david-mears-2:
Not a real PR, just a place to collect some Rich comments on a wip branch


scripts/run-dependencies
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+set -exo pipefail
+
+HERE=$(realpath "$(dirname $0)")
+
+# Build and run the database
+# TODO: Make build step skippable
+"$HERE"/../db/scripts/build
```
richfitz, line 9: As the db image won't change that much (because the schema is the bit that changes and not baked into the image) we might build this elsewhere. leave it this way for now perhaps and we can easily split it out later

richfitz (commented):
Success after:

* installing the correct version of node (.nvmrc might make this easier)
* getting the correct path to the script

The good news is I can see both your R and db pages working correctly.

There are some surprising (to me) pauses in bringing things up - the schema application took ages but perhaps that was installing packages? After running up the web app it seems to take a few seconds for everything to be ready. And I don't get the globe rendered but see this in the logs:

```
[warn] [Vue warn]: Failed to resolve component: Globe
If this is a native custom element, make sure to exclude it from component resolution via compilerOptions.isCustomElement.
[warn] [Vue warn]: Component <Anonymous> is missing template or render function.
```

PR #21, opened 2024-08-27
Refactor: Use NuxtLink component instead of CNavLink

david-mears-2:
This allows pre-fetching of links (by default this happens when links become visible).

https://nuxt.com/docs/api/components/nuxt-link
https://medium.com/@pepcorns/deep-dive-into-the-nuxt-link-component-b9e189866991

The prefetching feature defaults to being on: https://nuxt.com/docs/api/nuxt-config#defaults

Configuring the prefetch trigger to be 'interaction' instead of visibility requires upgrading the Nuxt version so I updated all dependencies while I was at it.


EmmaLRussell (approved):
LGTM!

PR #25, opened 2024-09-06
Make run button functional

david-mears-2:
* Add a functional submit button and all the related gubbins: api endpoint and handler, tests.
* Use a pinia store for largeScreen

You can test this locally by seeing if clicking the run button takes you to a page showing you the dummy run id.


components/ParameterForm.vue
```diff
@@ -118,27 +126,36 @@ const renderAsRadios = (parameter: Parameter) => {
...
-  }
+const submitForm = async () => {
+  if (!formData.value) {
+    return;
+  };
+
+  formSubmitting.value = true;
+  const response = await $fetch<NewScenarioData>("/api/scenarios", {
+    method: "POST",
+    query: { // Using query instead of body because I couldn't work out how to send a body in the integration test.
```
absternator, line 139: should defs use body here... your code shouldn't change to accommodate the tests...

components/ParameterForm.vue
```diff
@@ -118,27 +126,36 @@ const renderAsRadios = (parameter: Parameter) => {
...
 onMounted(() => {
-  setFieldSizes();
-  window.addEventListener("resize", setFieldSizes);
-});
-onBeforeUnmount(() => {
-  window.removeEventListener("resize", setFieldSizes);
+  pageMounted.value = true;
 });
+
+const appStore = useAppStore();
```
absternator, line 157: should move this at top with all variable definitions... nice to keep nicely separated with methods

components/ParameterForm.vue
```diff
@@ -118,27 +126,36 @@ const renderAsRadios = (parameter: Parameter) => {
...
-const largeScreen = ref(true);
-const breakpoint = 992; // CoreUI's "lg" breakpoint
-const setFieldSizes = () => {
-  if (window.innerWidth < breakpoint) {
-    largeScreen.value = false;
-  } else {
-    largeScreen.value = true;
-  }
+const submitForm = async () => {
+  if (!formData.value) {
```
absternator, line 132: when does this happen?

components/ParameterForm.vue
```diff
@@ -118,27 +126,36 @@ const renderAsRadios = (parameter: Parameter) => {
...
+  };
+
+  formSubmitting.value = true;
+  const response = await $fetch<NewScenarioData>("/api/scenarios", {
+    method: "POST",
+    query: { // Using query instead of body because I couldn't work out how to send a body in the integration test.
+      parameters: formData.value,
+    },
+  }).catch((error: FetchError) => {
+    console.error(error);
```
absternator, line 143: display this error back to user

components/ParameterForm.vue
```diff
@@ -118,27 +126,36 @@ const renderAsRadios = (parameter: Parameter) => {
...
+  });
+
+  if (response) {
+    const { runId } = response;
+    navigateToData.value = `/scenarios/${runId}`;
```
absternator, line 148:
```suggestion
   await navigateTo(`/scenarios/${runId}`)
```

unless reason for assignment before?

components/ParameterForm.vue
```diff
@@ -1,9 +1,11 @@
 <template>
-  <div>
+  <div v-show="pageMounted">
```
absternator, line 2: is this for playwright again? i feel its causing another render now... probs best to take the L in the playwright test rather than decrease app performance and increase bloat

components/ParameterForm.vue
```diff
@@ -65,13 +69,14 @@
       <CButton
         id="run-button"
         color="primary"
-        :size="largeScreen ? 'lg' : undefined"
+        :size="screenIsLarge ? 'lg' : undefined"
         type="submit"
+        :disabled="formSubmitting || !pageMounted"
```
absternator, line 74: a loading icon could be nice also whilst `formSubmitting`

components/SideBar.vue
```diff
@@ -1,12 +1,12 @@
...
     :visible="visible"
-    :unfoldable="largeScreen"
-    :overlaid="largeScreen"
+    :unfoldable="screenIsLarge"
+    :overlaid="screenIsLarge"
     class="sidebar-fixed border-end"
     @hide="handleHide"
   >
-    <CSidebarNav role="navigation">
+    <CSidebarNav role="navigation" data-testid="sidebar">
```
absternator, line 9: is data-testid neccesary? can it be tartged without this in tests? i am just against data-testid because its something users cant see and imaginary of sorts... and tests should mimic what users have access too

server/api/scenarios.post.ts
```diff
@@ -0,0 +1,17 @@
+import { runScenario } from "@/server/handlers/scenarios";
+import { defineRApiEventHandler } from "~/server/utils/defineRApiEventHandler";
+import type { NewScenarioResponse } from "@/types/apiResponseTypes";
+import type { ParameterDict } from "@/types/apiRequestTypes";
+
+export default defineRApiEventHandler(
+  async (event): Promise<NewScenarioResponse> => {
+    const query = getQuery(event);
```
absternator, line 8: as per prev comment should be reading body not query params

server/api/scenarios.post.ts
```diff
@@ -0,0 +1,17 @@
...
+
+    const modelParameters = JSON.parse(query.parameters as string) as ParameterDict;
+
+    // Delegate to runScenario so that the logic can be unit-tested.
```
absternator, line 12: dont need this comment

server/handlers/scenarios.ts
```diff
@@ -0,0 +1,28 @@
+import type { EventHandlerRequest, H3Event } from "h3";
+import { fetchRApi } from "@/server/utils/rApi";
+import type { NewScenarioData, NewScenarioResponse } from "@/types/apiResponseTypes";
+import type { ParameterDict } from "@/types/apiRequestTypes";
+
+const rApiNewScenarioEndpoint = "/scenario/run";
+
+// Send a request to the R API for a job run.
```
absternator, line 8: dont neeed comments again.. the code you wrote is clean and understandable

stores/appStore.ts
```diff
@@ -0,0 +1,21 @@
...
+  getters: {
+    screenIsLarge: (state: AppState): boolean => {
+      return state.largeScreen;
+    },
+  },
+  actions: {
+    initializeAppState(): void {
+      this.largeScreen = true;
+    },
+    setScreenSize(bool: boolean): void {
```
absternator, line 17: should this be `setScreenSizeToLarge` .. also function param should be descrptive .. eg. isLargeScreen rather than bool

absternator (commented):
Nice work !!! looks clean and works great!!!!!! just left some minor comments

components/ParameterForm.vue
david-mears-2: I don't have a reason to prefer body over query; do you?

components/ParameterForm.vue
david-mears-2: When does !formData.value be undefined, you mean? It would only happen if the inputs were nullish, which shouldn't happen, but possibly could with some combination of form autofilling and hydration lifecycle race conditions? It's defensive coding.

components/ParameterForm.vue
david-mears-2: This is for testing only, because I couldn't get a spy on navigateTo.

components/ParameterForm.vue
david-mears-2: Another render?

components/ParameterForm.vue
david-mears-2: Yep, I put a spinner into the button. Not that it hangs around long enough for a human to see it!

components/SideBar.vue
david-mears-2: It's a handle for playwright to grab onto. Playwright has a limited range of locators, but it does also have 'getByRole' that we could use here - except that we have two things with a navigation role so it's not unique. Also, when the navbar is narrow, it has no text that we could have done getByText on.

stores/appStore.ts
david-mears-2: Good points.

components/ParameterForm.vue
david-mears-2: Making a ticket

david-mears-2: Context - I only expect there to be errors here in development, since there is no known way for the form data to fail validation at the moment (you can't empty the input fields)

components/ParameterForm.vue
david-mears-2: Let me know if you manage to reproduce the re-rendering because I can't make it happen for me -- would be great to have steps to reproduce.

components/ParameterForm.vue
absternator: Semantically the use of query params is incorrect here... they are more for filtering, sorting / picking certain ids... when doing post request and sending data its semantically correct to use a body

components/ParameterForm.vue
absternator: sweet

components/ParameterForm.vue
absternator: nah you are good i just thought it would

components/ParameterForm.vue
david-mears-2: (maybe it will though when the subsequent page takes actual time to load)

server/api/scenarios.post.ts
david-mears-2: I've moved this comment - which was getting duplicated for each api endpoint - into a new .md "API design decisions.md"

stores/appStore.ts
david-mears-2: I have changed the function param to be called large, otherwise kept this the same

components/ParameterForm.vue
david-mears-2: I tried for some time to use the nuxt test utils to pass through a body, but there didn't seem to be a way.
To be clear, the R API is still receiving a body, it's only the web app that's taking a query parameter.

tests/unit/components/ParameterForm.spec.ts
```diff
@@ -147,12 +132,46 @@ describe("parameter form", () => {
...
+    });
+
+    const buttonEl = component.find("button[type='submit']");
+    expect(buttonEl.attributes("disabled")).not.toBe("");
+    await buttonEl.trigger("click");
+    expect(buttonEl.attributes("disabled")).toBe("");
+
+    await flushPromises();
+    // I couldn't find a way to spy on the navigateTo function, so this is a second-best test that we
+    // at least construct the correct path to navigate to.
```
EmmaLRussell, line 168: Did you try stubGlobal or mockNuxtImport?
https://stackoverflow.com/questions/77827093/how-to-check-if-navigateto-was-called-in-nuxt-js

EmmaLRussell (commented):
Great stuff, good to get the store in. 

Seems like some of the testing was a headache, which I'm probably adding to with some of these comments.. I think it would be really good to avoid some of these test-related-code-polluting additions if possible. Have made a couple of suggestions which you may have already tried. Happy to have a look at any of these if you're sick of bashing against them yourself - can't guarantee to get anywhere, as this is mostly all new to me too, but I think it's worth solving these now if we can, and can avoid proliferation.

components/ParameterForm.vue
david-mears-2: I tried again today with a fresh brain and am making headway :)

components/ParameterForm.vue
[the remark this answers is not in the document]
david-mears-2: Luckily, this problem disappears [next PR](https://github.com/jameel-institute/daedalus-web-app/pull/27), because by then we can wait for version data as our signal that everything is mounted.

components/ParameterForm.vue
david-mears-2: I had the comment in the test file - there are two other places it could be, here, and by the data attribute itself.

david-mears-2: I think our only option besides this is just not to test the form submission as part of component tests. That means we would rely only on e2e tests for the form submission. But since our e2e test doesn't mock the R API, we can't verify that the correct parameters were passed to it by the component.

stores/appStore.ts
```diff
@@ -0,0 +1,21 @@
...
+  }),
+  getters: {
+    screenIsLarge: (state: AppState): boolean => {
+      return state.largeScreen;
+    },
+  },
+  actions: {
+    initializeAppState(): void {
+      this.largeScreen = true;
+    },
```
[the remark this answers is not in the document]
david-mears-2: True. Previously this had more going on!

components/ParameterForm.vue
```diff
@@ -118,26 +128,32 @@ const renderAsRadios = (parameter: Parameter) => {
...
-const largeScreen = ref(true);
-const breakpoint = 992; // CoreUI's "lg" breakpoint
-const setFieldSizes = () => {
-  if (window.innerWidth < breakpoint) {
-    largeScreen.value = false;
-  } else {
-    largeScreen.value = true;
-  }
+  formSubmitting.value = true;
+  const response = await $fetch<NewScenarioData>("/api/scenarios", {
```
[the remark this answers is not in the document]
david-mears-2: I was imagining all of the scenario-related endpoints living under a single 'scenarios' namespace, rather than varying between 'scenario' and 'scenarios' depending on the use case. I believe in CRUD it is conventional to standardize on the plural form in route names.

components/SideBar.vue
david-mears-2: It happens that this is vestigial anyway.

layouts/default.vue
```diff
@@ -23,6 +23,28 @@ const sidebarVisible = ref(false);
 function handleToggleSidebarVisibility() {
   sidebarVisible.value = !sidebarVisible.value;
 }
+
+const appStore = useAppStore(useNuxtApp().$pinia);
+appStore.initializeAppState();
+
+const setScreenSize = () => {
+  // As this function uses window, it can only be run client-side, so we have to pass the $pinia instance
```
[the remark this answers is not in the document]
david-mears-2: I went through an adventure since writing this, submitted a PR to clarify the docs of the nuxt pinia module, and eventually the maintainer let me know that in fact "you don't need to pass the pinia instance within setup"
https://github.com/vuejs/pinia/pull/2766

As such, the next PR removes this, here: https://github.com/jameel-institute/daedalus-web-app/pull/27/files#diff-433a9abd4ca70520f69fc064c160bb093918ce26dda6087a8bbfdbe2895d1944

tests/e2e/runAnalysis.spec.ts
```diff
@@ -6,16 +6,25 @@ test.beforeAll(async () => {
...
   await expect(page.getByText("Simulate a new scenario")).toBeVisible();
+  await expect(page.getByRole("form")).toBeVisible({ timeout: 30000 });
 
   await page.selectOption('select[id="pathogen"]', { label: "Influenza 1957" });
   await page.selectOption('select[id="response"]', { label: "Elimination" });
   await page.selectOption('select[id="country"]', { label: "United States" });
-  await page.click('div[aria-label="Advance vaccine investment"] label[for="medium"]');
+  await page.click('div[aria-label="Advance vaccine investment"] label[for="low"]');
+
+  await expect(page.getByRole("form")).toHaveAttribute("data-test-form-data", "{\"country\":\"United States\",\"pathogen\":\"influenza-1957\",\"response\":\"elimination\",\"vaccine\":\"low\"}");
```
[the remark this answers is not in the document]
david-mears-2: We can be confident that the values of the inputs are what we just selected. I think the test to do for the e2e would be to wait until the next page is implemented, and test that the displayed parameters are correct.

tests/integration/rApi.spec.ts
```diff
@@ -122,4 +120,49 @@ describe("endpoints which consume the R API", { sequential: true }, async () =>
...
   });
+
+  // In these tests, Mockoon is configured to check the request body for all the expected parameters (and respond with the
+  // appropriate status code etc.), as a way to test that the parameters (and model version) are being passed through all the way
+  // to the R API.
+  describe("post api/scenarios", async () => {
+    it("returns a successful response when the mock server responds successfully", async () => {
+      const queryString = `parameters={"mockoonResponse":"successful","country":"Thailand","pathogen":"sars-cov-1","response":"no_closure","vaccine":"none"}`;
+
+      const response = await nuxtTestUtilsFetch(`/api/scenarios?${queryString}`, { method: "POST" });
```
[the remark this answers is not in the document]
david-mears-2: My problem had been that I was trying to pass objects into the 'body' property that just got lost in the post, as it were. I eventually solved this by finding that I could wrap the parameters in a `FormData` object, and decompose that at the other end.

tests/unit/components/ParameterForm.spec.ts
david-mears-2: Yes.

components/ParameterForm.vue
```diff
@@ -1,9 +1,11 @@
 <template>
-  <div>
+  <div v-show="pageMounted">
     <CForm
       v-if="props.metadata && formData"
       class="inputs"
-      :data-test="JSON.stringify(formData)"
+      role="form"
+      :data-test-form-data="JSON.stringify(formData)"
```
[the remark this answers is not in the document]
david-mears-2: Removed the dependency on this from tests. In the component test, I think the test relying on this was redundant anyway.

components/ParameterForm.vue
```diff
@@ -36,6 +38,7 @@
                 autocomplete="off"
                 :label="option.label"
                 :value="option.id"
+                :disabled="!pageMounted"
```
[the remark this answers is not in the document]
david-mears-2: Removed in next PR.

components/ParameterForm.vue
```diff
@@ -49,7 +52,8 @@
             :id="parameter.id"
             v-model="formData[parameter.id]"
             :aria-label="parameter.label"
-            class="form-select" :class="[largeScreen ? 'form-select-lg' : '']"
+            class="form-select" :class="[screenIsLarge ? 'form-select-lg' : '']"
+            :disabled="!pageMounted"
```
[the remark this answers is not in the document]
david-mears-2: Removed in next PR.

components/ParameterForm.vue
david-mears-2: Well, until such time as it can test the result page, I suppose.

stores/appStore.ts
```diff
@@ -0,0 +1,21 @@
...
+    setScreenSize(large: boolean): void {
+      this.largeScreen = large;
+    },
+  },
```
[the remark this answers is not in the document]
david-mears-2: Good to know.

components/ParameterForm.vue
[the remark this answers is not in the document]
EmmaLRussell: Yeah, it's a bit uneasy. Fair enough!

tests/e2e/runAnalysis.spec.ts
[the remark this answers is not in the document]
EmmaLRussell: Hang on, that doesn't look like the line (I thought) this was a comment for anymore, I'm pretty sure this was about the stringified JSON - or I thought it was!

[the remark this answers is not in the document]
david-mears-2: Still looks like it's commenting on the stringified JSON to me!

tests/unit/components/ParameterForm.spec.ts
EmmaLRussell: OK, let's not hold up this branch, but it would be really good to find a way to mock global methods, and it seems really weird if the alleged mocking capacities of vitest/nuxt test utils aren't going to let us to that. Tbh if that proves impossible personally I'd rather not clutter up the code with extra props for the sake of unit tests, as long as the navigation etc gets tested in e2e tests. These components are going to be busy and confusing enough as it is, without a bunch of extra stuff that doesn't really need to be there.

components/ParameterForm.vue
```diff
@@ -135,12 +134,15 @@ const submitForm = async () => {
     return;
   };
 
+  const formDataObject = new FormData();
+  Object.entries(formData.value).forEach(([key, value]) => {
+    formDataObject.append(key, value.toString());
+  });
```
EmmaLRussell, line 140: I'd assumed that we were just going to post JSON in the same format that the R API accepts and pass it straight through? You're saying that doesn't work (in the tests), that ordinary objects get turned into... what? Completely empty body?

It's not something to do with needing to set the content type is it? Does it default to POSTing form encoded data so it barfs on plain objects but accepts FormData? Hm, the ofetch docs suggest it should deal with that transparently and add the header itself, but it is supicious... Could try setting that header explicitly?

EmmaLRussell (commented):
I think at some point we're going to have to work out how to POST JSON with Nuxt tests, as it really ought to be possible 😆 It's not terrible to post form values as form data but it would be nice to have the option not to.. 

See previous comment on the mocking navigate controversy - I'd rather take out the props that are just there for unit tests if it's really not possible to mock, and make sure the navigration is tested in e2e. I might have a fiddle around with the mocking, maybe I'll stumble on some other way of doing it... but that can go in another branch.

tests/e2e/runAnalysis.spec.ts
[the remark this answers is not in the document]
EmmaLRussell: Sorry I'm just blind as a :bat:

EmmaLRussell:
> See previous comment on the mocking navigate controversy - I'd rather take out the props that are just there for unit tests if it's really not possible to mock, and make sure the navigration is tested in e2e. I might have a fiddle around with the mocking, maybe I'll stumble on some other way of doing it... but that can go in another branch.

This seemed to work for me: https://github.com/jameel-institute/daedalus-web-app/pull/29/files#

components/ParameterForm.vue
david-mears-2: Emma found this solution: https://github.com/jameel-institute/daedalus-web-app/pull/29

components/ParameterForm.vue
david-mears-2: I have found something that appears to be working - JSON stringifying the body (in the test). The oftech docs claimed that this happened automatically, but evidently not?

david-mears-2: See commit 6530321 for solution

david-mears-2: It seems that still it behaves differently under test than in real life, so I had to add in a check for whether the body was being received as an object (as in real life, and in e2e tests) or a JSON parsable string (as in the integration test). I hope this is tolerable, as the price for moving away from FormData.

EmmaLRussell: As discussed. maybe try adding header `Content-Type: application/json` when doing the POST.

EmmaLRussell (approved):
OK this one's cooked as far as I'm concerned!

PR #26, opened 2024-09-06
Use pinia store for version data

david-mears-2:
Very piecemeal, I hope you like tiny PRs! But some large thinking about it...

For the case of version data, we don't need to make everything (page rendering, loading...) wait around for the R API to respond. So I've configured this initialization of store data to be very lazy. By that I mean it's supposed not to block the execution of JS (I avoid using 'await') or slow down client-side rendering.

That said, I can see that the version data is included in the version of the page that is rendered server-side (by using 'inspect source' in a browser or 'payload' in Nuxt devtools). That implies to me that, by the time the server finishes rendering the page, the fetch has already resolved (the R API responded quickly enough), and there was time enough for this data to be included in the render.

I thought the lazyness would add value in situations where the R API responds slowly for whatever reason, during the server-side render. In that case, the lazyness would mean the server renders a page that doesn't include any version data. I believe the client, on receiving this, would retry the request, and (since we set 'lazy: true' - [docs here](https://nuxt.com/docs/api/composables/use-fetch#params)) would not slow down any page navigation in order to do so.

That's the theory, anyway, but when I set 'timeout: 10000' in server/utils/rApi.ts (i.e. tell the web app to give up on the R API after 10 seconds), and simulate infinitely long response times from the R API using Mockoon, the page does in fact hang for 10 seconds on refresh. As if the `lazy: true` option is being ignored. (Was using a page that doesn't make any metadata requests to isolate the test case.) Then I thought, is it because I'm doing a manual refresh rather than following a link? That's a possible reading of the documentation. But `lazy: true` doesn't seem to be necessary / make a difference with that, either, in the sense that whether or not I use that option, _following a link_ always takes me nicely to the next page, doing the desired fallback of omitting the version data in the sidebar, as I would have hoped would happen on the initial load of the website. (Side note - to do that test of the lazy:true option required moving the sidebar component from the default layout into the page being loaded, otherwise Nuxt just re-uses the existing layout when you navigate to the next page, skipping the request for version data.) 

So I don't have confidence that the commit that tries to introduce lazyness is having any of the desired effect.

So I'm not sure whether to keep the second commit in, in case it does provide value that I haven't been able to prove, or to revert to the first commit, so that at least we can be sure we know what's going on / we don't have any code that only illusorily does what we think it does.

I'm leaning to just revert to the first commit, so that we always remain abreast of how things work, but interested in your thoughts.

### DIFF
--- a/components/SideBar.vue
+++ b/components/SideBar.vue
@@ -57,22 +57,20 @@
 
 <script lang="ts" setup>
 import { CIcon } from "@coreui/icons-vue";
-import type { VersionData } from "@/types/apiResponseTypes";
 
-const { data: versionData } = useFetch("/api/versions") as { data: Ref<VersionData> };
+const appStore = useAppStore();
+const { screenIsLarge, getVersions } = storeToRefs(appStore);
 
 const versionTooltipContent = computed(() => {
-  if (versionData.value) {
-    return `Model version: ${versionData.value.daedalusModel} \nR API version: ${versionData.value.daedalusApi} \nWeb app version: ${versionData.value.daedalusWebApp}`;
+  const vers = getVersions.value;
+  if (vers) {
+    return `Model version: ${vers.daedalusModel} \nR API version: ${vers.daedalusApi} \nWeb app version: ${vers.daedalusWebApp}`;
   } else {
     return undefined;
   }
 });
 
 const visible = defineModel("visible", { type: Boolean, required: true });
-
-const appStore = useAppStore();
-const { screenIsLarge } = storeToRefs(appStore);
 
 const resetSidebarPerScreenSize = () => {
   visible.value = screenIsLarge.value;
--- a/layouts/default.vue
+++ b/layouts/default.vue
@@ -25,7 +25,7 @@ function handleToggleSidebarVisibility() {
 }
 
 const appStore = useAppStore();
-await appStore.initializeAppState();
+appStore.initializeAppState();
 
 const setScreenSize = () => {
   const breakpoint = 992; // CoreUI's "lg" breakpoint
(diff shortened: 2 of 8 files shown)


david-mears-2:
One answer to consider is back out the second commit and set the timeout tolerance of the R API handler to something rather low. I should check at some point if our requests to it can be parallelised - 200ms here and there can add up! (I can think of the result page, which to load would want to fetch the versions, metadata, and results in three requests.) I already looked into caching on the web app side recently, but had some trouble (it played havoc with testing since there was no easy way to purge the cache.)

david-mears-2:
Plan for testing that parallelisation hypothesis in the last comment:

- Load version data synchronously (without 'await'), with a 2000ms timeout on the web app side, and a 1500ms delay on the mock R API side, a dozen times when app store initialises.
- See if Nuxt server-side rendering cleverly makes all these requests simultaneously, or if it takes 12 x 1500ms to render.
- See if Nuxt client-side rendering is similarly clever.
- If not, cry quietly, and then try running in prod mode in case that has better performance optimization.

david-mears-2:
Simple fix for now that appears to work is to move the trigger for requesting the version data out of the app store initialisation, and into an 'onMounted'.

david-mears-2:
I haven't done the test I planned [here](https://github.com/jameel-institute/daedalus-web-app/pull/26#issuecomment-2338477458), but might want to do it in order to do metadata and results requests simultaneously. But that said, given that the metadata was probably loaded on a previous page load than the navigation to the results page, there might not be much call for this optimization anyway, and it would just be a case of researching something that we really ought to know about how Nuxt works.

PR #32, opened 2024-09-12
jidea-76 run redis and api worker in dependencies

EmmaLRussell:
This branch includes the additional docker containers required for rrq support in the dependencies scripts. 

Currently targeting [this branch](https://github.com/jameel-institute/daedalus.api/pull/8) of the API. When you're happy with this PR, I'll merge the api branch, and point the scripts here back to main before merging. The deploy tool will be temporarily broken (hopefully very temporarily), but I think it's just going to be more straightforward to target main for all containers when I do the updates there..

The additional containers are:
- redis - rrq uses a redis db to manage the queue
- daedalus api worker - the worker runs the same image as the main api container, but with a different entry point which watches for tasks to run
- daedalus api configure - the queue needs to be configured once before any workers or controllers (task senders) attach. This is done with the same image with yet another entry point. This container is ephemeral and removes itself once the configuration is done. 

You should find that you can run and tear down dependencies using the scripts as normal, but should see additional containers `daedalus-redis` and `daedalus-api-worker` are run. The app should work as normal. 

I had to add a couple of little fixes for issues currently in main as a result of metadata updates 
- form can't assume all parameters will have options
- update to e2e test for new vaccine investment label.  

After this is merged, you'll be getting real responses from the run and status endpoint, but still pre-canned responses from the results endpoint.

### DIFF
--- a/scripts/run-dependencies
+++ b/scripts/run-dependencies
@@ -49,7 +49,7 @@ VOLUME=daedalus-results
 docker network create $NETWORK
 docker network connect daedalus daedalus-web-app-db
 
-docker volume create VOLUME
+docker volume create $VOLUME
 
 # Pull and run the R API image
 # TODO: reset this to latest when the rrq branch of the api is merged
--- a/scripts/run-dependencies
+++ b/scripts/run-dependencies
@@ -43,10 +43,51 @@ fi
 
 docker exec daedalus-web-app-db wait-for-db
 
-docker network create daedalus
+NETWORK=daedalus
+VOLUME=daedalus-results
+
+docker network create $NETWORK
 docker network connect daedalus daedalus-web-app-db
 
+docker volume create VOLUME
+
 # Pull and run the R API image
-R_API_IMAGE='mrcide/daedalus.api:latest'
+# TODO: reset this to latest when the rrq branch of the api is merged
+R_API_IMAGE='mrcide/daedalus.api:jidea-74'
 docker pull $R_API_IMAGE # There will be a 'latest' tag: https://github.com/jameel-institute/daedalus.api/pull/2#issuecomment-2252491603
-docker run -d --name daedalus-api --rm --network=daedalus -p 8001:8001 $R_API_IMAGE
+
+QUEUE_ID=daedalus-api-queue-$(uuidgen)
+
+docker run --rm -d \
+  -p 127.0.0.1:6379:6379 \
+  --name=daedalus-redis \
+  --network=$NETWORK \
+  redis
+
+# Ephemeral container for one-time queue configuration
+docker run --rm \
+  --network=$NETWORK \
+  --env=DAEDALUS_QUEUE_ID=$QUEUE_ID \
+  --env=REDIS_CONTAINER_NAME=daedalus-redis \
+  --entrypoint="/usr/local/bin/daedalus.api.configure" \
+  $R_API_IMAGE
+
+docker run --rm -d \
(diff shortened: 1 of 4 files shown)


scripts/run-dependencies
```diff
@@ -43,10 +43,51 @@ fi
...
+docker volume create VOLUME
+
 # Pull and run the R API image
-R_API_IMAGE='mrcide/daedalus.api:latest'
+# TODO: reset this to latest when the rrq branch of the api is merged
+R_API_IMAGE='mrcide/daedalus.api:jidea-74'
 docker pull $R_API_IMAGE # There will be a 'latest' tag: https://github.com/jameel-institute/daedalus.api/pull/2#issuecomment-2252491603
-docker run -d --name daedalus-api --rm --network=daedalus -p 8001:8001 $R_API_IMAGE
+
+QUEUE_ID=daedalus-api-queue-$(uuidgen)
```
EmmaLRussell, line 59: We could use the same queue id every time, but that can cause problems if the volume doesn't get removed cleanly for some reason, so it's generally safer for the dev/test scripts to generate a new id on every run.

david-mears-2: So is it the case that we have a single queue at any given time?

scripts/run-dependencies
```diff
@@ -43,10 +43,51 @@ fi
...
+
+docker run --rm -d \
+  -p 127.0.0.1:6379:6379 \
+  --name=daedalus-redis \
+  --network=$NETWORK \
+  redis
+
+# Ephemeral container for one-time queue configuration
```
david-mears-2, line 67: What makes it ephemeral? Does it get used once and hang around or do we remove it?

scripts/run-dev-dependencies
```diff
@@ -6,6 +6,9 @@ function cleanup() {
   set +ex # allow teardown errors
   docker kill daedalus-web-app-db
   docker kill daedalus-api
+  docker kill daedalus-api-worker
+  docker kill daedalus-redis
```
david-mears-2, line 10: Should the ephemeral container be killed here?

david-mears-2:
Note to self: I have locally rebased this on the next branch along -- jidea-57-non-country-inputs-rebasing-onto-front-end-brach -- maybe helpful for future merges

scripts/run-dependencies
EmmaLRussell: We should do, yes. But I've had cases of old volumes hanging around and causing problems, and Rich advised that it's standard in the test scripts to just generate a new queue id for yourself every time.

scripts/run-dependencies
```diff
@@ -43,10 +43,51 @@ fi
...
 docker exec daedalus-web-app-db wait-for-db
 
-docker network create daedalus
+NETWORK=daedalus
+VOLUME=daedalus-results
+
+docker network create $NETWORK
 docker network connect daedalus daedalus-web-app-db
 
+docker volume create VOLUME
```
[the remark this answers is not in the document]
EmmaLRussell: Oops! Yeah, I'm surprised that didn't fail!

scripts/run-dependencies
EmmaLRussell: Its entry point script is just doing one thing and then exiting so it stops  naturally  - unlike the api and the worker which wait to be given things to do. The `--rm` flag on the docker run means that it removes itself on exit.

david-mears-2 (approved):

PR #33, opened 2024-09-13
Jidea 57 mock navigate to 2

EmmaLRussell:
OK, this was very mysterious. The culprit for why this suddently stopped working appears to have been the h3 `readBody` in the test request handler. That was one of the few changes in the relevant files so I tried taking it out and just manually parsing the body, and the mock started working again. ..

I don't really understand why this made a difference - maybe something in that readBody was failing? But I tried debugging it, and it looked like it was getting the runId and doing the navigateTo OK in the component.  Maybe some sort of async issue?

So, yay, but also, huh?

### DIFF
--- a/components/ParameterForm.vue
+++ b/components/ParameterForm.vue
@@ -4,7 +4,6 @@
       v-if="props.metadata && formData"
       class="inputs"
       role="form"
-      :data-test-navigate-to="navigateToData"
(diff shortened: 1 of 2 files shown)


david-mears-2:
Well discovered... Perhaps we did not pour enough libations to the gods?

PR #34, opened 2024-09-17
Show params on results page

david-mears-2:
The unit of changes here encompasses:

* Removing breadcrumbs code - we can reimplement if and when we get time for it.
* Store parameter selection in the store, play selections back to the user on the result page.
* Allow navigating from the parameter selection playback to editing the inputs.

I'm still excluding the Globe component from branches, since that ticket is on hold, but you'll see a few things relating to it, like the store storing an 'globe: { interactive: false }' property - that one changes whether the globe is in the foreground, colored, and clickable, vs greyed out in the background and static.


components/ParameterForm.vue
```diff
@@ -148,13 +148,14 @@ const appStore = useAppStore();
...
+  // Initialize formData as a dictionary of parameters with values set to defaults if available,
+  // or to the previous scenario's values if any.
   paramMetadata.value?.reduce((acc, { id, defaultOption, options, updateNumericFrom }) => {
-    if (options && !updateNumericFrom) { // Excludes fields whose values depend on others'; we'll set these once all the defaults are set.
+    if (!acc[id] && options && !updateNumericFrom) { // Excludes fields whose values depend on others'; we'll set these once all the defaults are set.
       acc[id] = (defaultOption || options[0].id).toString();
     }
     return acc;
-  }, {} as { [key: string]: string }),
+  }, { ...appStore.currentScenario?.parameters } as { [key: string]: string }),
```
EmmaLRussell, line 158: I think this two-stage initialisation (with conditional based on the accumulator we're currently accumulating) is quite difficult to read. 
Can't we just initialise fully from the current scenario if there is one, rather than getting the reduce involved at the level of each parameter? I'd also suggest pulling out the initialise from defaults into a separate method to make the basic initialisation logic clearer
```
const initialiseFormDataFromDefaults = () => {
  return paramMetadata.value?.reduce((acc, { id, defaultOption, options, updateNumericFrom }) => {
    if (options && !updateNumericFrom) { // Excludes fields whose values depend on others'; we'll set these once all the defaults are set.
      acc[id] = (defaultOption || options[0].id).toString();
    }
    return acc;
  }, {} as { [key: string]: string })
}
const formData = ref(
  // Initialize formData as a dictionary of parameters with values set to defaults if available,
  // or to the previous scenario's values if any.
  appStore.currentScenario ? {...appStore.currentScenario.parameters} : initialiseFormDataFromDefaults()
);
```

components/ParameterForm.vue
```diff
@@ -235,10 +236,11 @@ const invalidFields = computed(() => {
   return invalids;
 });
 
+// Since some defaults depend on the values of other fields, this function should not be used to initialize form values.
 const defaultValue = (param: Parameter) => {
   if (param.updateNumericFrom) {
     const dependedOnParamId = param.updateNumericFrom.parameterId;
-    return param.updateNumericFrom.values[formData.value![dependedOnParamId]].default.toString();
+    return param.updateNumericFrom.values[formData.value[dependedOnParamId]]?.default?.toString();
```
EmmaLRussell, line 243: Sorry, I can't really tell what's prompted this change or what it's actually doing, and the comment doesn't really help. 

Is it using the default of the parameter this parameter depends on, if it's not already set..? So that would be happen if we're trying to get default hospital capacity, but we haven't initialised country yet?

This logic of this line is very tangled - I think splitting it up onto a couple of lines would help a lot to show what the intermediate values you're using are, and why.

components/ParameterForm.vue
```diff
@@ -235,10 +236,11 @@ const invalidFields = computed(() => {
   return invalids;
 });
 
+// Since some defaults depend on the values of other fields, this function should not be used to initialize form values.
```
EmmaLRussell, line 239: So what should it be used for? :confused:

components/ParameterForm.vue
```diff
@@ -297,20 +300,20 @@ const submitForm = async () => {
...
-const resetAllDependents = () => {
-  paramMetadata.value?.filter((param) => {
-    return param.updateNumericFrom !== undefined;
-  }).forEach(resetParam);
-};
+paramMetadata.value?.filter((param) => {
+  const isDependent = param.updateNumericFrom !== undefined;
+  const shouldBeSetByStore = appStore.currentScenario?.parameters?.[param.id];
+  return isDependent && !shouldBeSetByStore;
+}).forEach(resetParam);
```
EmmaLRussell, line 313: Why has this been moved to no longer happen in onMounted?

EmmaLRussell (commented):
Looks great! Now that you've proven it looks good with the flag, it might be best to hide the flag until we can show an accurate one - or at least change the default to UK flag for tomorrow's demo. 

Good call on the breadcrumbs. At some point we'll want to do something similar to remove the sidebar menu items which aren't going to be functional for MVP - maybe just hide them based on a feature switch rather than remove completely, so they're easy to reinstate. Could do the same thing with the breadcrumbs if you like. 

Bit confused about some of the parameter logic! 🤯

components/ParameterForm.vue
david-mears-2: Originally it ran in onMounted because it relies on metadata having been fetched. Then I wondered does it really need to be there, and tried moving it outside of onMounted, and letting Mockoon give us a delayed response (1000ms) to the metadata API call to see if it broke anything. That was fine, so I concluded we could run this in the setup script.

However, I don't think moving it out of 'onMounted' gains us much, and I'm happy to move it back.

components/ParameterForm.vue
david-mears-2: I think this change was prompted by the situation with Thailand -- something like, the country value was being defaulted to Thailand, but since that wasn't a country, there was no corresponding updateNumericFrom values? -- and so "Checking whether a field value is already at its default" (something done at render time) threw an error preventing the page from loading. So the addition of the question marks covers us (allows the form to still load) when we can't find any default value. But maybe we want the dramatic failure of a 500 page to alert us to the problem.

Happy to refactor to be clearer, anyhow.

david-mears-2:
```ts
if (param.updateNumericFrom) {
    const dependedOnParamId = param.updateNumericFrom.parameterId;
    const dependedOnValue = formData.value[dependedOnParamId];
    const dependentDefaultValue = param.updateNumericFrom.values[dependedOnValue]?.default;
    return dependentDefaultValue?.toString();
```

components/ParameterForm.vue
david-mears-2: Yep, that works, thanks!

david-mears-2:
Changed hard-coded flag to UK flag.

components/ParameterForm.vue
EmmaLRussell: Thanks for explanation - yeah, that makes sense I think. That situation shouldn't really happen, but I guess it's good to be somewhat resilient!

components/ParameterForm.vue
EmmaLRussell: So the parameter form itself isn't loaded until metadata is available anyway? If so, it shouldn't really make a difference - don't mind either way.

EmmaLRussell (approved):

PR #36, opened 2024-09-27
Jidea 97 show timeseries

david-mears-2:
Changes included here:

* Added TimeSeries and TimeSeriesLegend components, with visual testing
* TimeSeries is capable of zooming and syncing, and resizing if you hide another time series
* ParameterForm adapted to be able to be used in a modal
...


assets/icons/index.js
```diff
@@ -1,6 +1,201 @@
 import {
+  cifAd,
```
david-mears-2, line 2: This adds all the country flag icons.

PR #38, opened 2024-10-02
JIDEA-97: Show timeseries

david-mears-2:
I've done my best to exclude things that aren't directly relevant. It's not quite as many lines as it looks: 1000 are from storing the sample result response and 300 are from Mockoon recording responses.

Includes:
* Added TimeSeries and TimeSeriesLegend components, with visual testing
* TimeSeries is capable of zooming and syncing, and resizing if you hide another time series


components/TimeSeries.vue
```diff
@@ -0,0 +1,353 @@
...
+const data = computed(() => {
+  return appStore.timeSeriesData![props.seriesId].map((value, index) => [index + 1, value]);
+});
+const seriesMetadata = computed((): DisplayInfo | undefined => {
+  return appStore.metadata?.results?.time_series.find(({ id }) => id === props.seriesId);
+});
+const usePlotLines = props.seriesId === "hospitalised"; // https://mrc-ide.myjetbrains.com/youtrack/issue/JIDEA-118/
+const usePlotBands = props.seriesId === "hospitalised" || props.seriesId === "infect"; // https://mrc-ide.myjetbrains.com/youtrack/issue/JIDEA-118/
+// On zooming, the y-axis automatically rescales to the data (ignoring the plotLines). We want the plotLines
+// to remain visible, so we limit the y-axis' ability to rescale, by defining a minimum range.
```
EmmaLRussell, line 89:
```suggestion
// On zooming, the y-axis automatically rescales to the data (ignoring the plotLines). We want the plotLines
// to remain visible, so we limit the y-axis' ability to rescale, by defining a minimum range. This way the plotLines remain visible even when the maximum data value is less than the maximum plotline value.
```
Took me a couple of goes to understand this!

components/TimeSeries.vue
```diff
@@ -0,0 +1,353 @@
...
+    bands.push({ from: start, to: end, color: plotBandsColor, label: { text } });
+  });
+
+  return bands;
+});
+
+const allCharts = computed(() => {
+  // Excludes undefineds. When a chart is destroyed, Highcharts.charts retains its index, but the value is undefined.
+  return Array.from(Highcharts.charts).filter(chart => chart) as (Highcharts.Chart)[];
+});
```
EmmaLRussell, line 143: So this gets all charts loaded anywhere in the page? Again, this component doesn't feel like quite the right place for any code that's dealing with all charts. Could this go in a `TimeSeriesList` component?
Also, this might be a bit dangerous if we potentially end up using Highcharts for other elements in the page. Could we get to all charts by id or ref?

components/TimeSeries.vue
```diff
@@ -0,0 +1,353 @@
...
+    allCharts.value.forEach(({ series }) => {
+      // Get the point with the same x as the hovered point
+      const point = series[0].getValidPoints().find(({ x }) => x === chart.hoverPoint!.x);
+
+      if (point && point !== chart.hoverPoint) {
+        point.onMouseOver();
+      }
+    });
+  };
+}, 100, { leading: true });
```
EmmaLRussell, line 167: Is every TimeSeries doing this for every chart...?

components/TimeSeries.vue
```diff
@@ -0,0 +1,353 @@
...
+      },
+      buttons: {
+        contextButton: {
+          height: 20,
+          width: 22,
+          symbolSize: 12,
+          symbolY: 10,
+          symbolX: 12,
+          symbolStrokeWidth: 2,
+          // Omit 'printChart' and 'viewData' from menu items
```
EmmaLRussell, line 237: I guess `viewData` is limited usefulness but someone might want to print..?

pages/scenarios/[id].vue
```diff
@@ -87,6 +87,60 @@
...
+            class="card-body p-0"
+            @mouseleave="onMouseLeaveTimeSeries"
+            @mouseover="() => { hideTimeSeriesTooltips = false }"
+          >
+            <TimeSeries
+              v-for="(_, seriesId, index) in appStore.timeSeriesData"
+              :key="seriesId"
+              :series-id="seriesId"
+              :index="index"
+              :opened-accordions="openedTimeSeriesAccordions"
```
Comment (EmmaLRussell) [inline pages/scenarios/[id].vue:136]: As suggested in `TimeSeries`, this could be something like
```suggestion
              :open="openedTimeSeriesAccordions.includes(seriesId)"
```

pages/scenarios/[id].vue
```diff
@@ -96,13 +150,63 @@ import type { Parameter } from "~/types/parameterTypes";
...
+});
+
 onMounted(() => {
   appStore.globe.interactive = false;
+
+  if (!appStore.currentScenario.status.data?.done || appStore.currentScenario.result.data) {
+    statusInterval = setInterval(loadScenarioStatus, 200); // Poll for status every N ms
+    setTimeout(() => {
+      clearInterval(statusInterval); // Terminate polling for status after several seconds
+    }, 5000);
```
Comment (EmmaLRussell) [inline pages/scenarios/[id].vue:204]: This should probably result in some error being printed to screen about the model run timing out (if the interval hadn't already been cleared). Is 5s long enough

stores/appStore.ts
```diff
@@ -15,12 +32,52 @@ export const useAppStore = defineStore("app", {
...
+        status: scenarioStatusFetchStatus,
+        error: scenarioStatusFetchError,
+      } = await useFetch(`/api/scenarios/${this.currentScenario.runId}/status`) as {
+        data: Ref<ScenarioStatusData>
+        status: Ref<AsyncDataRequestStatus>
+        error: Ref<FetchError>
+      };
+
+      this.currentScenario.status = {
+        data: { ...scenarioStatusData.value, runId: undefined },
```
EmmaLRussell, line 60: Why set the `runId` to undefined here? Because you don't need it, because it's already on `currentScenario`? Could always just discard it - or could check it against currentScenario before setting the status.

tests/e2e/runAnalysis.spec.ts
```diff
@@ -53,6 +53,25 @@ test("Can request a scenario analysis run", async ({ page, baseURL, isMobile })
   await expect(page.getByText("Medium").nth(parameterLocatorIndex)).toBeVisible();
   await expect(page.getByText("200000").nth(parameterLocatorIndex)).toBeVisible();
 
+  // To regenerate these screenshots:
+  // 1. Insert a generous timeout so that screenshots are of the final chart, not the chart half-way through
+  //    its initialization animation: `await page.waitForTimeout(30000);`
+  // 2. Run the test with this flag: `npm run test:e2e -- --update-snapshots`
```
EmmaLRussell, line 59: Nice! :sunglasses:

EmmaLRussell (commented):
Looks great! 

Tiny quibble: on mobiles, the parameters, costs placeholder and timeseries are all glued together:
image
Would look nicer with a little bit of margin. RHS of Parameters and Time series aren't quite flush on either view, would look a little better if they were aligned. 

Were you going to find a place to include the description for the time series? Maybe on hover over the title?

I feel like there's an argument for an intermediate `TimeSeriesList` component to manage some aspects related to the coordination of the list of charts (chart height, etc), rather than leaving that up to each individual Time series. See inline comments. 

Looks like there's still some build failures...

stores/appStore.ts
david-mears-2: Yeah, this is an attempt to discard it, because it seems untidy/duplicatey to be storing multiple copies of the runId within the store - I feel we should refer to one source of truth.

The `ScenarioStatusData` interface marks runId as an optional parameter, since the interface is used to type the response from the R API.

david-mears-2:
I like the idea of the intermediate `TimeSeriesList` component and I'll give it a go.

pages/scenarios/[id].vue
Comment (david-mears-2) [reply pages/scenarios/[id].vue:169]: I updated this behavior to be as you see in this screen recording:

[Screencast from 06-10-24 23:46:41.webm](https://github.com/user-attachments/assets/7e4939c0-f1a2-4202-a2dd-af138eefee62)

Namely, it gives up after 15 seconds, but warns that things are taking longer than expected at 5 seconds.

For future reference, you can configure Mockoon to induce this behaviour in the app (by mocking the R API scenario status endpoint to say 'pending' forever), using the below Mockoon config:

[mockoon.json](https://github.com/user-attachments/files/17271149/mockoon.json)

components/TimeSeries.vue
david-mears-2: I omitted printChart because whenever you print, there is a mouseover, which induces a tooltip, which is included in the print preview, which is bad, and likely not worth the effort to fix.

image

david-mears-2: Downloading the PDF, PNG, JPEG, or SVG should suffice for any printing needs!

components/TimeSeries.vue
```diff
@@ -0,0 +1,353 @@
...
+  "--cui-accordion-bg": "rgba(255, 255, 255, 0.7)",
+};
+// We need each chart to have a higher z-index than the next one so that the exporting context menu is always on top and clickable.
+// Also, they should be at least 3 so that they are above .accordion-button:focus
+const zIndex = (Object.keys(appStore.timeSeriesData!).length - props.index) + 3;
+const yUnits = props.seriesId === "dead" ? "deaths" : "cases"; // TODO: Make this depend on a 'units' property in metadata. https://mrc-ide.myjetbrains.com/youtrack/issue/JIDEA-117/
+const minAccordionHeight = 150;
+const minTotalAccordionHeight = 500;
+
+const isOpen = computed(() => props.openedAccordions.includes(props.seriesId));
```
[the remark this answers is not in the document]
david-mears-2: This *was* a bit of efficiency because we were already passing openedAccordions in order to use openedAccordions.length. But now all that is getting handled in the intermediate component TimeSeriesList.

components/TimeSeries.vue
david-mears-2: Yes, whenever any time series component's (singular) chart has a relevant mouse/touch event, we find the point that is hovered, work out what the equivalent points on the other charts are, and trigger those points on the other charts to show a tooltip.

components/TimeSeries.vue
david-mears-2: Good point on 'dangerous'! Now we store the charts as a dict belonging to the TImeSeriesList component.

david-mears-2:
> Tiny quibble: on mobiles, the parameters, costs placeholder and timeseries are all glued together:
Would look nicer with a little bit of margin. RHS of Parameters and Time series aren't quite flush on either view, would look a little better if they were aligned.

These will be fixed in the more styling-ish changes that were excluded from this PR for the sake of brevity.
 
> Were you going to find a place to include the description for the time series? Maybe on hover over the title?

Well caught, it seems I lost these when I moved to using accordions - now reinstated. They show on hovering over an icon. Not quite sure if the (?) icon I'm using is preferable to an [i] icon ([this one](https://coreui.io/icons/linear/info/)).

david-mears-2:
I have no theory for what was going on with the screenshots - where I was seeing a differently shaped chart, at a different scale of numbers, for firefox only - except for "something to do with how using playwright with the '--update-screenshots' flag only updates screenshots if they are significantly different from the existing screenshot". However, blowing up the directory of screenshots and creating them ACTUALLY anew - i.e. without using the '--update-screenshots' flag - brought all of the charts in line with each other again.

components/TimeSeries.vue
```diff
@@ -0,0 +1,323 @@
...
+      buttons: {
+        contextButton: {
+          height: 20,
+          width: 22,
+          symbolSize: 12,
+          symbolY: 10,
+          symbolX: 12,
+          symbolStrokeWidth: 2,
+          // Omit 'printChart' and 'viewData' from menu items
+          menuItems: ["printChart", "downloadCSV", "downloadXLS", "separator", "downloadPNG", "downloadJPEG", "downloadPDF", "downloadSVG", "separator", "viewFullscreen"],
```
EmmaLRussell, line 221: Ah, you've added printChart back in - I think it's fine to leave it out as you said, if you get unwanted tooltip. Either way, the comment above should be consistent with the list.

components/TimeSeriesList.vue
```diff
@@ -0,0 +1,116 @@
...
+    return Math.max(minTotalAccordionHeight, (Object.keys(appStore.timeSeriesData!).length * minAccordionHeight));
+  } else {
+    return minTotalAccordionHeight;
+  }
+});
+// Share available height equally between open accordions. Avoid division by zero.
+const accordionHeight = computed(() => openedAccordions.value.length ? (maxTotalAccordionHeight.value / openedAccordions.value.length) : 1);
+const chartHeightPx = computed(() => Math.min(accordionHeight.value, maxAccordionHeight) - (2 * accordionBodyYPadding));
+
+const storeChart = (seriesId: string, chart: Highcharts.Chart) => {
```
EmmaLRussell, line 49: Because `charts` is fully managed in this parent component, and does not interact with the pinia store, I feel like `storeChart`/`unstoreChart` is potentially a bit misleading. How about `chartCreated`/`chartDestroyed` or `chartAdded`/`chartRemoved`?

EmmaLRussell (approved):
Thanks, this looks really clean now. Couple of tiny comments, but happy to approve.

components/TimeSeries.vue
david-mears-2: Oh, that was an accident.

PR #42, opened 2024-10-07
Use modal instead of page navigation to edit parameter form

david-mears-2:
Also included:

* Small fixes excluded from last PR for readability - see commit messages for descriptions of these
* Rename scenarios/\[id\] page

I recommend that the easiest way to read this PR would be a commit at a time, since the changes in each commit aren't really related to each other, with the exception of the e2e tests commit, which relates to the foregoing changes.

Janky screen recording of phone-sized device using the modal in portrait and in landscape:

[Screencast from 07-10-24 17:31:26.webm](https://github.com/user-attachments/assets/a56a2ffe-d696-4ded-ac03-17686ab8f2e8)


pages/scenarios/[runId].vue
```diff
@@ -13,8 +13,42 @@
...
+              <template #toggler="{ togglerId, on }">
+                <span :aria-describedby="togglerId" v-on="on">
+                  <CButton color="light">
+                    Parameters
+                  </CButton>
+                  <CIcon icon="cilPencil" class="form-icon link-secondary" />
+                </span>
+              </template>
+            </CTooltip>
+          </div>
```
Comment (EmmaLRussell) [inline pages/scenarios/[runId].vue:51]: There's a bit of duplication here with the Parameters control markup underneath for large screen. Could pull out the functional bits  (including the modal) into an `EditParameters` component and include it in the right parent/with the right styling here depending on screen size. Also helps keep the sprawl in this page under control a bit.

components/ParameterForm.vue
```diff
@@ -142,6 +142,10 @@ import type { Parameter, ParameterSet, ValueData } from "@/types/parameterTypes"
 import { TypeOfParameter } from "@/types/parameterTypes";
 import type { NewScenarioData } from "@/types/apiResponseTypes";
 
+const props = defineProps<{
+  inModal?: boolean
+}>();
```
EmmaLRussell, line 147: I think it would be fine to make this required, and include isModal=false for the instance in new scenario page.

tests/unit/components/ParameterForm.spec.ts
```diff
@@ -133,12 +150,16 @@ describe("parameter form", () => {
...
+      expect(rangeInput.element.value).toBe("17000");
+    }, 500);
 
     await shortList.find("input[value='no']").setChecked();
-    expect(numericInput.element.value).toBe("2000");
-    expect(rangeInput.element.value).toBe("2000");
+    setTimeout(() => {
+      expect(numericInput.element.value).toBe("2000");
+      expect(rangeInput.element.value).toBe("2000");
+    }, 500);
```
EmmaLRussell, line 162: I'm surprised you need these `setTimeouts` - nextTick should do the trick? I'm also a bit confused about what shortList represents here - this is a use case we don't currently have where button group param potentially has dependent params to update?

pages/scenarios/[runId].vue
Comment (david-mears-2) [reply pages/scenarios/[runId].vue:51]: True

tests/unit/components/ParameterForm.spec.ts
david-mears-2: nextTick works great, somehow I forgot about it.

pages/scenarios/[runId].vue
Comment (david-mears-2) [reply pages/scenarios/[runId].vue]: Made the suggested component

EmmaLRussell (approved):
LGTM

PR #46, opened 2024-10-10
Update npm packages

david-mears-2:
Was doing this to see if it fixed a bug - it didn't - but we can still use the updates.

Since I updated the linter's version, there are some linting changes, mainly reordering of imports.

Note that locally you'll need to do `npx playwright install` for next time you run e2e tests.


EmmaLRussell (approved):
We can always choose if we really want to follow the latest default diktats of the linter which might spare churn like this, but this looks good to me.

PR #47, opened 2024-10-10
Remove hard timeout for polling for scenario status/results

david-mears-2:
~~Also, since I could not figure out the bug relating to the component test and mocking pinia, and time is of the essence, I wrote an e2e test instead.~~

For bracket's sake!! :anger:  I spent definitely over an hour looking for this yesterday, and today realised that the mistake preventing me from using pinia in my test was this change:

correct:
`    const component = await mountSuspended(ScenariosIdPage, { global: { stubs, plugins } });`
incorrect:
`    const component = await mountSuspended(ScenariosIdPage, { global: { stubs }, plugins })`;

Now that I've written the e2e test anyway... do you reckon it's worth keeping it? It's configured to run for only one browser, since it takes 20-30 seconds.


types/parameterTypes.ts
```diff
@@ -28,9 +28,9 @@ export interface Parameter {
   id: string
```
david-mears-2, line 28: Changes to this file are to bring types in line with actual outputs of the R API, which returns 'undefined's (or absences of properties) rather than 'null's.

pages/scenarios/[runId].vue
```diff
@@ -120,56 +121,82 @@ import type { Parameter } from "~/types/parameterTypes";
...
       }
     }, 5000);
+    // Some runs take an especially long time, e.g. Singapore + Omicron.
     setTimeout(() => {
-      // If the job isn't completed within 10 seconds, terminate polling for status.
       if (!appStore.currentScenario.status.data?.done) {
-        jobTakingLongTime.value = false;
-        stoppedPolling.value = true;
+        jobSlow.value = false;
+        jobReallySlow.value = true;
```
Comment (EmmaLRussell) [inline pages/scenarios/[runId].vue:198]: Logic might be a little simpler if you just let slow keep being true when reallySlow is also true

pages/scenarios/[runId].vue
```diff
@@ -120,56 +121,82 @@ import type { Parameter } from "~/types/parameterTypes";
...
   }
 };
 
+// Use useAsyncData to store the time once, during server-side rendering: avoids client render re-writing value.
+const { data: timeOfFirstStatusPoll } = await useAsyncData<number>("timeOfFirstStatusPoll", async () => {
+  return new Promise<number>((resolve) => {
+    resolve(new Date().getTime());
+  });
+});
+
```
Comment (EmmaLRussell) [inline pages/scenarios/[runId].vue:153]: Maybe just try declaring and setting directly in the setup tag - I *think* it should only get calculated once, on the server, but may be wrong!
`const timeOfFirstStatusPoll = new Date().getTime();`

pages/scenarios/[runId].vue
```diff
@@ -120,56 +121,82 @@ import type { Parameter } from "~/types/parameterTypes";
...
 
 onMounted(() => {
   appStore.globe.interactive = false;
 
   if (!appStore.currentScenario.status.data?.done && appStore.currentScenario.runId) {
-    statusInterval = setInterval(loadScenarioStatus, 200); // Poll for status every N ms
+    pollForStatusEveryNSeconds(0.2);
     setTimeout(() => {
-      // If the job isn't completed within five seconds, give user the information about the run status.
+      // If the job isn't completed within M seconds, give user the information about the run status.
```
Comment (EmmaLRussell) [inline pages/scenarios/[runId].vue:189]:
```suggestion
      // If the job isn't completed within 5 seconds, give user the information about the run status.
```

pages/scenarios/[runId].vue
```diff
@@ -120,56 +121,82 @@ import type { Parameter } from "~/types/parameterTypes";
...
       if (appStore.currentScenario.status.data?.runStatus !== runStatus.Complete) {
```
Comment (EmmaLRussell) [inline pages/scenarios/[runId].vue:190]: Not from this branch, but I just noticed that the two timeouts use different ways to tell if a job is finished: 
`appStore.currentScenario.status.data?.runStatus !== runStatus.Complete` vs `!appStore.currentScenario.status.data?.done` - I think the latter is more reliable because status can be in error, but the job is still done.

EmmaLRussell (commented):
I think it's nice to have the e2e test for now, and we can always remove it when (if?) the model improves such that the scenario runs without triggering the "longer than expected" view.

pages/scenarios/[runId].vue
```diff
@@ -65,18 +65,19 @@
...
-        The analysis is taking longer than expected. Please
-        <NuxtLink prefetch-on="interaction" to="/scenarios/new">
-          <span>try again</span>
-        </NuxtLink>.
+    <CAlert
+      v-if="!appStore.timeSeriesData && (jobSlow || jobReallySlow) && appStore.currentScenario.status.data?.runStatus"
+      :color="jobReallySlow ? 'warning' : 'info'"
+    >
+      <p v-if="jobReallySlow">
+        The analysis is taking longer than expected. Please wait or try again.
```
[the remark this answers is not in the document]
Comment (david-mears-2) [reply pages/scenarios/[runId].vue:74]: We got some feedback - I think from Singapore - that it was surprising that 'try again' didn't re-run anything but dropped you back at the start. I like your rephrasing.

pages/scenarios/[runId].vue
Comment (david-mears-2) [reply pages/scenarios/[runId].vue:155]: So what happens is this:

image

Here you see logs from one initial load of the 'new' page, as when you refresh that page or navigate to it from another site. This is rendered server side first, and then the setup script is run again on the client-side.

Most of the time (and indeed ALL of the time until we add the ability to refresh the results page), people will navigate to the results page from another page rather than loading it directly or refreshing it. And in _those_ cases, the setup script is run only on the client side:

image

Finally, here are some logs of when we use useAsyncData when doing an initial load of the 'new' page. You can see that Nuxt retains the original value.
image

You can also inspect that value from the 'payload' section of the Nuxt DevTools:

image

or by doing "View page source" from the browser (i.e. the originally loaded HTML, not the current HTML):

image

pages/scenarios/[runId].vue
```diff
@@ -65,18 +65,19 @@
...
       </p>
-    </CAlert>
-    <CAlert v-else-if="jobTakingLongTime && appStore.currentScenario.status.data?.runStatus" color="info">
-      <p class="mb-0">
+      <p>
         Analysis status: {{ appStore.currentScenario.status.data?.runStatus }}
       </p>
+      <p class="mb-0">
+        Time elapsed: {{ secondsSinceFirstStatusPoll }} seconds
+      </p>
```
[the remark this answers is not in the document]
Comment (david-mears-2) [reply pages/scenarios/[runId].vue:81]: I think it's useful to know how long you've been waiting for! I'll rephrase.

pages/scenarios/[runId].vue
Comment (david-mears-2) [reply pages/scenarios/[runId].vue:191]: -> 'a few'

EmmaLRussell (approved):

PR #49, opened 2024-10-10
Fix tooltip widths

david-mears-2:
# Before

image

# After

Screenshot from 2024-10-14 11-29-19


assets/scss/_variables.scss
```diff
@@ -3,6 +3,7 @@ $app-header-margin-bottom: 1.5rem;
 $min-wrapper-height: calc(100dvh - $app-header-height - $app-header-margin-bottom);
 $container-padding: 1.5rem;
 $sidebar-narrow-width: 4rem;
+$iphone-se-width: 375px;
```
EmmaLRussell, line 6: So you're using this as the narrowest expected screen width? 
Might be clearer to renamed this to something that reflects that e.g. `$minimum-screen-width` and include a comment about what the current narrowest supported model is..?

EmmaLRussell (commented):
Looks good!

assets/scss/_variables.scss
david-mears-2: Reasonable!!!

EmmaLRussell (approved):

PR #51, opened 2024-10-15
JIDEA-105 Costs pie

david-mears-2:
Adding a display of the costs results in the form of a pie chart (or, technically speaking, a 'sunburst').


components/CostsCard.vue
```diff
@@ -0,0 +1,290 @@
...
+  }
+  return {
+    height: maxSizeInTopRight,
+    right: rightPosition,
+    top: 0,
+  };
+});
+
+// The second of three candidate styles that will be evaluated for how much space they can provide for the pie.
+const belowUsdTotalStyle = computed(() => {
```
EmmaLRussell, line 116: It might be a bit less confusing to call this `belowTotals` rather than `belowUsdTotal` - as it's slotting the pie entirely under the totals, rather than sneaking it into the corner between the two totals as bottomRight does. it's kind of irrelevant for this layout which total is lowest, as you calculate from `containers.total`

components/CostsCard.vue
```diff
@@ -0,0 +1,290 @@
...
+    rightPosition = (containers.cardBody.width.value - maxSizeInBottomRight) - containers.usd.width.value;
+  }
+  return {
+    height: maxSizeInBottomRight,
+    right: rightPosition,
+    bottom: 0,
+  };
+});
+
+const styleWithLargestPie = computed(() => [topRightStyle.value, belowUsdTotalStyle.value, bottomRightStyle.value].sort((a, b) => b.height - a.height)[0]);
```
EmmaLRussell, line 150: Or could be something like
```suggestion
const styleWithLargestPie = computed(() => Math.max(...[topRightStyle.value, belowUsdTotalStyle.value, bottomRightStyle.value].map((s) => s.height));
```
..but just because I don't love reading sort methods!

components/CostsCard.vue
```diff
@@ -0,0 +1,290 @@
...
+              </span>
+            </span>
+          </p>
+          <p class="mt-0">
+            [insert note about GDP basis]
+          </p>
+        </div>
+      </div>
+      <CostsPie
+        v-if="pieStyle.height"
```
EmmaLRussell, line 54: Could 0 height happen? I tried making the window unreasonably tiny and the totals and pie just fell off the bottom of the card, but the pie was still shown. 
Might be worth setting a minimum height for the card?

components/CostsCard.vue
```diff
@@ -0,0 +1,290 @@
...
+        :hide-tooltips="hideTooltips"
+        :pie-size="styleWithLargestPie.height"
+        :style="pieStyle"
+        @mouseleave="onMouseLeavePie"
+        @mouseover="() => { hideTooltips = false }"
```
EmmaLRussell, line 59: What's the need for the tooltip management here? Doesn't the library manage showing and hiding tooltips? Maybe a comment on this?

components/CostsPie.vue
```diff
@@ -0,0 +1,261 @@
...
+        // When the clicked chart 'point' is in the second level of the sunburst,
+        // that means some drilling down or up (traversing up and down the chart) is triggered.
+        const isDrilling = event.point.node.level !== 1 && event.point.node.children.length > 0; // No drill-down if no children to drill down to;
+        if (!isDrilling) {
+          return;
+        }
+        const isDrillingUp = event.point.drillId === appStore.totalCost?.id;
+        const isDrillingDown = isDrilling && !isDrillingUp;
+        chart.update({ series: { levels: chartLevelsOptions(isDrillingDown) } });
+      },
```
EmmaLRussell, line 100: Is this an override of the chart's drill down behaviour or was there no default implementation?

components/CostsPie.vue
```diff
@@ -0,0 +1,261 @@
...
+    },
+    dataLabels: {
+      rotation: 0,
+      rotationMode: "auto", // Without this, labels sometimes appear in the top left at random. https://github.com/highcharts/highcharts/issues/18953
+      format: "{point.name}",
+      style: {
+        fontSize: "1.1rem", // TODO: Make font-size smaller on small screens (or disappear on very small)
```
EmmaLRussell, line 107: Still planning on doing this TODO? One for another ticket?

components/CostsPie.vue
```diff
@@ -0,0 +1,261 @@
...
+      events: {
+        fullscreenOpen() {
+          this.update({ chart: { backgroundColor: chartBackgroundColorOnExporting } });
+        },
+        fullscreenClose() {
+          this.update({ chart: { backgroundColor: chartBackgroundColor } });
+        },
+      },
+      style: {
+        fontFamily: "ImperialSansText, sans-serif", // TODO: Make the font-family derive from a globally configurable constant
```
EmmaLRussell, line 139: ..and this one I guess

tests/unit/components/CostsCard.spec.ts
```diff
@@ -0,0 +1,220 @@
...
+
+  if (costsPiePosition === "top-right-corner") {
+    /*
+
+    [t][t][ ][p][p][p][p][ ]
+    [g][g][ ][p][p][p][p][ ]
+    [u][ ][ ][p][p][p][p][ ]
+    [ ][ ][ ][p][p][p][p][ ]
+
+    */
```
EmmaLRussell, line 81: Great stuff, thanks for this.

tests/unit/components/CostsPie.spec.ts
```diff
@@ -0,0 +1,171 @@
...
+    getOptions: actual.getOptions,
+    chart: () => mockChart,
+    charts: actual.charts,
+    _modules: actual._modules,
+    win: actual.win,
+    Pointer: actual.Pointer,
+  };
+});
+
+describe("costs card", () => {
```
EmmaLRussell, line 59:
```suggestion
describe("costs pie", () => {
```

tests/unit/components/TimeSeries.spec.ts
```diff
@@ -53,6 +53,60 @@ vi.mock("highcharts", async (importOriginal) => {
 });
 
 describe("time series", () => {
+  it("should initialise the chart with the correct options", async () => {
```
EmmaLRussell, line 56: Was this just a missing test? Didn't think the time series behaviour had changed, other than removing intervention labels.

EmmaLRussell (commented):
Looks great! Comments are mostly questions.. 

I wonder if a lot of the complexity of figuring out the best layout to maximise pie size results from the totals taking up a squarish patch - and is that really required? A simpler approach might be to put GDP% and USD next to each other horizontally and always slot the pie underneath? That might alllow a greater use of the space? Maybe something to consider when you're looking at fixing the text size in a further ticket. 

Are the inner ring colors an average of the colors of their children? For GDP and Education it looks a bit odd because one child's color seems to be identical (or almost) to its parent while the other is much darker. 

It looks a bit odd how really large segments on the outer ring don't get labels, but I guess that decision is being made by the charting library based on length of label and horizontal space available. I think this will look less strange when the table view is added. 

Ooh, but if I drill down, then click the "Total" link I see more labels than I did before! some overlapping edges etc. Is this expected?
Screenshot from 2024-10-17 10-36-50


When the pie is positioned under the totals, it feels like the it's quite far away from the legend which pertains to it. Though it is neat to keep the legend in the card header...

absternator:
you probs need to pull in main

david-mears-2:
> Ooh, but if I drill down, then click the "Total" link I see more labels than I did before! some overlapping edges etc. Is this expected?

That's a bug :bug:  :cry: My intended behaviour was to show labels for the lowest level only when drilled down (otherwise the text can be too much)

components/CostsCard.vue
david-mears-2: Doesn't that return a number (a height) rather than a style? Or is it some fancy syntax I've not seen

absternator:
also feel as though we should get rid of snapshot testing... just ends us something we have to maintain that dosetn give us anything... most people will just update snapshot when it fails... thus becomes pretty useless form of testing

components/CostsPie.vue
david-mears-2: The intention here is to update the configuration such that the lowest level does or does not show its labels, depending on which level of drill-down we are currently on.

absternator:
should be able to see bottom border of the card enclosing the pie chart.. like the time series data

absternator:
image
image

getting these error on local when I refresh the page after already ran

david-mears-2:
Notes from chatting with Anmol and Emma:

1) clarify tooltips in costs pie to say e.g. "lifeyears: working-age adults" instead of just "working-age adults"
2) for this costs chart, and potentially time series too, give users a clue that hovering might provide more information.
3) mock up a different type of costs chart for the workshop presentation
4) one idea for making the division between the top three children more obvious: add spacing between the top-level delineations
5) try highcharts' built-in legend again

tests/unit/components/TimeSeries.spec.ts
david-mears-2: Yeah just backfilling

david-mears-2:
> Are the inner ring colors an average of the colors of their children? For GDP and Education it looks a bit odd because one child's color seems to be identical (or almost) to its parent while the other is much darker.

Looked for a way to configure this but I don't believe one exists.

david-mears-2:
> It looks a bit odd how really large segments on the outer ring don't get labels, but I guess that decision is being made by the charting library based on length of label and horizontal space available. I think this will look less strange when the table view is added.

Actually I deliberately switch off datalabels on the outer ring when there are three levels visible. Do you think that's suboptimal?
Screenshot from 2024-10-18 18-49-23
Screenshot from 2024-10-18 18-49-31

components/CostsCard.vue
david-mears-2: It keeps the tooltips up when you mouse off, unconfigurably it seems.

david-mears-2: Will comment

components/CostsPie.vue
david-mears-2: Another ticket

david-mears-2:
Since last reviews I've removed the logic about searching for the largest square to fit the pie in.

components/CostsCard.vue
```diff
@@ -0,0 +1,191 @@
...
+const onMouseLeavePie = () => {
+  setTimeout(() => {
+    hideTooltips.value = true;
+  }, 500);
+};
+
+// Fit the pie into the available width
+const pieSize = computed(() => {
+  return totalsContainerWidth.value * 0.6;
+});
```
EmmaLRussell, line 89: Now that the layout is fixed, would this not be simpler to do with css `width: 60%` in the `pieStyle`?

EmmaLRussell (commented):
Looks good! Legend seems more natural in its new position. One question below.

components/CostsCard.vue
david-mears-2: Good point. In any case, this is removed in Anmol's costs table PR, in favour of two fixed width options depending on `appStore.largeScreen`.

EmmaLRussell: Ah ok, will just approve here, if it's on its way out anyway.

EmmaLRussell (approved):

EmmaLRussell:
..oh pending the build being fixed!

PR #53, opened 2024-10-15
JIDEA-147: Display the headline cost as a percent of national GDP

david-mears-2:
Builds on #51 

We are using 2018 data for national GDPs (converted from daily GVAs by the API)

Also: Calculate percentages for costs pie tooltips too


david-mears-2:
Converting to draft until costs pie branch is resolved

PR #57, opened 2024-10-16
Jidea-83 Parameter help UI

EmmaLRussell:
This branch displays metadata descriptions for parameters and their options as help text in the Parameter Form. There are three types of help displayed:
1. Descriptions of parameters. These are shown as tooltips triggered by hover over a ? icon next to the Parameter title. I've made a new `ParameterHeader` component to include this, to avoid repetition of markup for each parameter type. 
2. Descriptions of radio button options. These are shown as tooltips triggered by hovering over the buttons. It does not appear to be possible to apply `CTooltip` directly to `CFormCheck` so I was obliged to wrap the form check in a surrounding element - however this caused styling problems as much of the styling of radio buttons assumes that the button will be a direct child of its parent button group, without the new intervening element. So I had to recreate some styles in the theme for the grandchild buttons... 
3. Descriptions of options in drop-down selects.  We currently have descriptions only for the "Response" parameter. These are displayed in a custom vue select control to allow templating of options - there are quite a few css changes related to this.  Using this control means we get option searchability for free which is nice, but see inline comment one nasty little gotcha relating to this, which meant that the value of the select could be unset (so we need to intervene to reset it again).. 

Notes following our discussion in the office:
- help icons are floated to the right for the parameter help. I haven't made the icons smaller, as they now feel less connected to the parameter name and its associated icon, and I don't want to make them too small to distinguish, but see what you think. 
- I've made the options more compact vertically and increased the height  of the opened dropdown. However, this does not clip itself nicely to the bottom of the viewport as the native select does, so I'm limiting this to 400px which should work for most screen sizes. (We could reproduce the native behaviour with some nasty js if you think it's worth it!)


david-mears-2:
For generating light blues that match well with existing blues, try using `--cui-primary-bg-subtle`, which is calculated based on what we assign to `$primary` in the variables.scss file.

components/ParameterForm.vue
```diff
@@ -168,6 +184,14 @@ const formData = ref(
   // or to the previous scenario's values if any.
   appStore.currentScenario.parameters ? { ...appStore.currentScenario.parameters } : initialiseFormDataFromDefaults(),
 );
+
+// Making the vue select searchable means that it's possible to unset a parameter value (to undefined) if you clear the search
+// input. In this case we just want to be able to revert to the previous value it had. However, this is tricky as we
+// don't get the previous value in any watch of formData since it's watching deep changes in an object, and we can't
+// use a computed setter for values in an object type. So here we keep a copy of the last full dictionary and reset in the watch
+// if required.
+const previousFullFormData = ref({ ...formData.value });
```
EmmaLRussell, line 193: This is obviously quite gross. Potentially avoidable by allowing the select drop down to be unset and using the invalids to show a validation error on submit - however that's not trivial since we're not using a core ui input, which gives all the nice validation Ui and styling. Could also just not make the select searchable, but that
does seem useful for country because there are so many options...

assets/scss/_theme.scss
```diff
@@ -125,3 +125,37 @@ body {
...
+.btn-group-lg > {
+  .radio-btn-container > {
+    .btn {
+      --cui-btn-padding-y: .5rem;
+      --cui-btn-padding-x: 1rem;
+      --cui-btn-font-size: 1.25rem;
+      --cui-btn-border-radius: var(--cui-border-radius-lg);
+    }
+  }
+}
```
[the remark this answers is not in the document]
david-mears-2: Was it related to the use of 'scoped' in an opening `<style lang="scss" scoped>` tag? This is sometimes helpful but as I add more styles to a component I end up dropping the 'scoped' because it doesn't capture elements that are not directly listed in the template (i.e. excludes elements that are parts of child/imported components), and I do my own scoping if necessary by giving the component a unique id or class that I can scope my styles to.

david-mears-2:
I'll review this fully later or tomorrow but I still think the help icon size could be smaller. When https://github.com/jameel-institute/daedalus-web-app/pull/63/ is merged, the class .smaller-icon will become available, which reduces size to 90%.

david-mears-2:
https://github.com/jameel-institute/daedalus-web-app/pull/63 is now merged

EmmaLRussell:
> I'll review this fully later or tomorrow but I still think the help icon size could be smaller. When #63 is merged, the class .smaller-icon will become available, which reduces size to 90%.

ok, will do

EmmaLRussell:
> > I'll review this fully later or tomorrow but I still think the help icon size could be smaller. When #63 is merged, the class .smaller-icon will become available, which reduces size to 90%.
> 
> ok, will do

done!

david-mears-2:
I think the population growth in the help tooltips is getting to be a bit cluttered, and at some point we should look for a different tidier way of displaying this info, e.g. showing the help text automatically above the input when a dropdown is opened. Or maybe all it needs is a different icon: a question mark without a circle might be sufficient to reduce the sense of clutter.

david-mears-2:
Before:

image

After:

image

Could you reset the padding/margin so these are all aligned?

assets/scss/_theme.scss
```diff
@@ -137,3 +137,47 @@ body {
...
+
+// In vue-select parameter dropdowns, avoid highlighting both focused (used when navigating with keyboard) and
+// hovered menu items, by forcing unhighlighted background colour on all menu items if they have a sibling which
+// is hovered, unless they are the unhovered item is currently selected (also highlighted, in a different colour).
```
david-mears-2, line 178: Some of words may been or perhaps added

components/ParameterForm.vue
david-mears-2: I wasn't able to make the select drop down set a parameter to unset when I removed this code, how can this be reproduced?

tests/unit/components/ParameterForm.spec.ts
```diff
@@ -1,9 +1,12 @@
 import ParameterForm from "@/components/ParameterForm.vue";
 import { emptyScenario, mockedMetadata, mockPinia, updatableNumericParameter } from "@/tests/unit/mocks/mockPinia";
-
+import { CButtonGroup, CTooltip } from "@coreui/vue";
 import { mockNuxtImport, mountSuspended, registerEndpoint } from "@nuxt/test-utils/runtime";
 import { flushPromises } from "@vue/test-utils";
 import { FetchError } from "ofetch";
+import { beforeEach, describe, expect, it, vi } from "vitest";
```
david-mears-2, line 7: We don't need to import anything from vitest ever - it's automatically imported.

david-mears-2:
```suggestion
```

tests/unit/components/ParameterForm.spec.ts
```diff
@@ -318,4 +344,20 @@ describe("parameter form", () => {
 
     expect(component.findComponent({ name: "CSpinner" }).exists()).toBe(true);
   });
+
+  it("form data reverts to previous id update includes falsy values", async () => {
```
david-mears-2, line 348: Which grammar does this be?

david-mears-2: Oh I see: id -> if

EmmaLRussell:
> I wasn't able to make the select drop down set a parameter to unset when I removed this code, how can this be reproduced?

If I recall: Click on the input, type a few letters as if searching, then delete your search text, then click away to unfocus the drop down.

EmmaLRussell:
> showing the help text automatically above the input when a dropdown is opened. 

The trouble of this is that you need to provide the real estate to provide that in - or have it in a popover type thing which would look weird. 

>Or maybe all it needs is a different icon: a question mark without a circle might be sufficient to reduce the sense of clutter.

Which is more subtle, but maybe too subtle... worth a try maybe though. 

I don't find the current view too cluttered, but maybe I'm just used to it! 

Could declutter by removing the parameter icons.. although they are cute!

assets/scss/_theme.scss
EmmaLRussell: ?? Is this a request to make this comment more comprehensible? It's a fair request!

david-mears-2: I couldn't grammar this
> unless they are the unhovered item is currently selected

david-mears-2:
Another idea could be to have one larger (?) in the top right of the parameter form, with a tooltip saying 'Click for an explanation of each of the parameters' or something.

EmmaLRussell:
> Could you reset the padding/margin so these are all aligned?

This was happening because .select-container's margins were relying on .row's slightly negative margins, which are overridden by our custom css in the scenario page when the row occurs inside a card - this sets margins to 0. It's an unintended consequence for this to affect the modal which is not visually "inside" the card - teleporting the modal out to the body fixes the problem.

components/ParameterForm.vue
```diff
@@ -390,4 +425,52 @@ onMounted(() => {
...
+.select-container {
+   margin-left: 0.7rem;
+   margin-right: 0.55rem;
+}
+
+.numeric-header {
+  padding-right: 2.2rem;
+}
+
+.vue-select {
```
absternator, line 438: should this vue selector styles be in globals? since any future selects should look the same

absternator (approved):
LGTM!!! just a small comment :smile:

components/ParameterForm.vue
EmmaLRussell: Ooh good point  - yes, possibly. But  I think these styles are quite specific to the context of the parameter form, so I think it's ok for them to stay here for now. If we ever do add any more vue selects, we'll split out what can go be shared in globals then I think.

PR #58, opened 2024-10-17
Switch off visual testing with Playwright

david-mears-2:
You'll like this one.

### DIFF
--- a/tests/e2e/downloadExcel.spec.ts
+++ b/tests/e2e/downloadExcel.spec.ts
@@ -22,6 +22,6 @@ test("can download Excel file for scenario results", async ({ page, baseURL }) =
   await download.path();
 
   // Filename is derived from parameter values, in this case the defaults
-  const expectedFilename = "daedalus_United Kingdom_sars_cov_1_none_none_26200.xlsx";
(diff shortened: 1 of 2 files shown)


david-mears-2:
Test failed on small screens since I was overly assertive in the test - I’ll just check that there is an axis not what the markers are

> Locator: locator('#prevalence-container .highcharts-xaxis-labels text').first()
    Expected string: "100"
    Received string: "250"

tests/e2e/downloadExcel.spec.ts
```diff
@@ -22,6 +22,6 @@ test("can download Excel file for scenario results", async ({ page, baseURL }) =
   await download.path();
 
   // Filename is derived from parameter values, in this case the defaults
-  const expectedFilename = "daedalus_United Kingdom_sars_cov_1_none_none_26200.xlsx";
+  const expectedFilename = "daedalus_Thailand_sars_cov_1_none_none_22000.xlsx";
```
EmmaLRussell, line 25: Could make this a pattern rather than fixed string, which allows the default hospital capacity to vary if data ever gets updated. Though I don't think that's going to be an issue for a while!

EmmaLRussell (commented):
Thanks! Still a couple of failures on mobile. Because the element in questions isn't in viewport on smaller screen?

PR #59, opened 2024-10-18
Jidea-151 Costs table

absternator:
The following PR is for work for displaying the costs table as well as with the pie chart. The following design decisions were made: 
- make table show first and then costs pie on right (visually more fitting)
- show numbers is millions (as some are too small for billions)
- fixed size of pie to 450px for large screens and 300 px for rest --> was the best after playing with all resolutions
- Not showing total as this is duplicate info from the total at the top

image
image
image
image


stores/appStore.ts
```diff
@@ -136,5 +136,11 @@ export const useAppStore = defineStore("app", {
         }
       })();
     },
+    getCostLabel(costId: string): string {
+      const name = this.metadata?.results.costs.find(
+        cost => cost.id === costId,
+      )?.label;
+      return name || costId;
+    },
```
david-mears-2, line 144: I like that you've moved this to the store.

tests/e2e/runAnalysis.spec.ts
```diff
@@ -41,17 +45,44 @@ test("Can request a scenario analysis run", async ({ page, baseURL, headless })
...
   // 2. Delete the screenshots directory, ./<this-file-name>-snapshots
   // 3. Run the tests with `npm run test:e2e` to regenerate the screenshots - tests will appear to fail the first time.
   // Make sure to stop any local development server first so that Playwright runs its own server, in production mode, so that the
   // Nuxt devtools are not present in the screenshots.
-  if (headless) {
-    await expect(page.locator(".accordion-body").first()).toHaveScreenshot("first-time-series.png", { maxDiffPixelRatio: 0.04, timeout: 15000 });
-    await expect(page.locator(".accordion-body").nth(1)).toHaveScreenshot("second-time-series.png", { maxDiffPixelRatio: 0.04 });
-    await expect(page.locator(".accordion-body").nth(2)).toHaveScreenshot("third-time-series.png", { maxDiffPixelRatio: 0.04 });
+  if (headless && useVisualScreenshotTesting) {
+    await expect(page.locator(".highcharts-background").first()).toHaveScreenshot("first-time-series.png", { maxDiffPixelRatio: 0.04, timeout: 15000 });
```
david-mears-2, line 83: I think these changes to the e2e test appear here because the branch is out of date with the latest version of jidea-105-14-oct

tests/unit/components/CostsCard.spec.ts
```diff
@@ -1,11 +1,9 @@
 import type { ScenarioResultData } from "@/types/apiResponseTypes";
-import type { VueWrapper } from "@vue/test-utils";
 import CostsCard from "@/components/CostsCard.vue";
 import { emptyScenario, mockPinia } from "@/tests/unit/mocks/mockPinia";
 import { mockResultResponseData } from "@/tests/unit/mocks/mockResultResponseData";
 import { mountSuspended } from "@nuxt/test-utils/runtime";
 import { waitFor } from "@testing-library/vue";
-import { mockResizeObserver } from "jsdom-testing-mocks";
```
david-mears-2, line 8: Could you also excise this from npm too please?

tests/unit/components/CostsPie.spec.ts
```diff
@@ -75,8 +76,8 @@ describe("costs card", () => {
 
   it("should populate the cost data into the chart when the data is loaded after the component is mounted", async () => {
     await mountSuspended(CostsPie, {
-      global: { stubs, plugins: [mockPinia()] },
-      props: { hideTooltips: false, pieSize: 100 },
+      global: { stubs, plugins: [mockPinia({ metadata: mockedMetadata }, true, false)] },
```
david-mears-2, line 79: You don't need to import and pass in the mockedMetadata -- mockPinia already does this by default (if you set  the second parameter to true)

tests/unit/mocks/mockPinia.ts
```diff
@@ -227,5 +227,5 @@ export const mockPinia = (appState: Partial<AppState> = {}, includeMetadata: boo
     },
   };
 
-  return createTestingPinia({ initialState, createSpy: vi.fn });
+  return createTestingPinia({ initialState, createSpy: vi.fn, stubActions });
```
david-mears-2, line 230: Another open PR, #47, adds the ability to stub actions from pinia in a bit more of a generic way, allowing the user of mockPinia to pass in a *dict* of options that is forwarded to createTestingPinia. Please could you copy the way this is done in the relevant commit: https://github.com/jameel-institute/daedalus-web-app/pull/47/commits/ae8adb3c1d060ecfe2017fa5c6417a5a507492ef

components/CostsTable.vue
```diff
@@ -0,0 +1,45 @@
...
+
+<script lang="ts" setup>
+const appStore = useAppStore();
+</script>
+
+<style scoped>
+.nested-row {
+  background-color: #f1f3f5;
+}
+</style>
```
david-mears-2, line 45:
```suggestion
<style scoped>
.nested-row {
  background-color: #f1f3f5;
}

td {
  padding-left: 0.5rem;
}
</style>
```

This looks a little neater to me on hover.

david-mears-2 (changes requested):
Looks good :)

A couple of changes and discussion points.

tests/unit/mocks/mockPinia.ts
david-mears-2: Though actually (as I'm fixing elsewhere) I think the parameter
```ts
  testingOpts: TestingOptions = {},
```
should be
```ts
  testingOpts: Partial<TestingOptions> = {},
```

components/CostsTable.vue
```diff
@@ -0,0 +1,45 @@
+<template>
+  <table v-if="appStore.totalCost" class="table rounded table-hover table-sm">
+    <thead class="border-bottom-2 border-black">
+      <tr>
+        <th>Loss</th>
+        <th>Total in M$</th>
```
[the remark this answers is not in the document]
absternator: yeah thought it was a bit confusing... but the usual table format for currency usually just numbers and can't use abbreviated as it differs and seems odd in a table..... I've changed to heading to **$ in millions** as that seems better

components/CostsTable.vue
```diff
@@ -0,0 +1,45 @@
+<template>
+  <table v-if="appStore.totalCost" class="table rounded table-hover table-sm">
+    <thead class="border-bottom-2 border-black">
+      <tr>
+        <th>Loss</th>
```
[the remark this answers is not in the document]
absternator: ive removed first column header and second column is as change above ($ in millions)...

david-mears-2:
A suggestion I liked from the scientists: omit the dollar sign on each row and just have it in the column header. This prevents confusion if your eye goes directly to the row without reading the unit in the column header.

david-mears-2:
For some reason, now this branch uses the costs pie colours in a different order than before (see screenshots). I prefer blue to be the color Highcharts tends to apply to the Life years segment, since that's almost always the largest.
Screenshot from 2024-10-22 13-08-56
Screenshot from 2024-10-22 13-09-20

components/CostsTable.vue
```diff
@@ -0,0 +1,45 @@
+<template>
+  <table v-if="appStore.totalCost" class="table rounded table-hover table-sm">
+    <thead class="border-bottom-2 border-black">
+      <tr>
+        <th />
+        <th>$ in millions</th>
```
david-mears-2, line 6:
```suggestion
        <th>$, millions</th>
```

components/CostsTable.vue
```diff
@@ -0,0 +1,45 @@
...
+          <td class="ps-2">
+            {{ appStore.getCostLabel(childCost.id) }}
+          </td>
+          <td>{{ formatCurrency(childCost.value) }}</td>
+        </tr>
+        <template v-if="childCost.children">
+          <tr
+            v-for="grandChildCost in childCost.children"
+            :key="grandChildCost.id"
+            class="nested-row"
```
david-mears-2, line 24:
```suggestion
            class="nested-row fw-lighter"
```

components/CostsTable.vue
david-mears-2: I think this suggestion failed to make it in. Also, a mild change to the font weight of the thead:

```suggestion
<style scoped>
.nested-row {
  background-color: #f1f3f5;
}

td {
  padding-left: 0.5rem;
}

thead tr th {
  font-weight: 600 !important;
}
</style>
```

david-mears-2 (commented):
See comment just now - Could you please try to work out what's causing the costs pie colours to change? The originals won't be the final colours but it would be really good if we were in complete control of the colours, so we need to understand how it was affected.

david-mears-2 (approved):
Thanks so much @absternator, I'm happy with this pending Emma's once-over.

EmmaLRussell (approved):
This looks great!

It looks a little bit odd, having some values presented as B or T USD, others as 100s of M USD, and VSL in Int'l$. But we're going to tidy that up post-workshop I think.

PR #62, opened 2024-10-18
J-IDEA 147: Express total cost in terms of % of national GDP

david-mears-2:
Builds on #51 and obseletes #53

We are using 2018 data for national GDPs (converted from daily GVAs by the API)

Also: Calculate percentages for costs pie tooltips too


components/utils/charts.ts
```diff
@@ -23,13 +23,12 @@ export enum LegendShape {
...
   const abbr = abbreviateMillionsDollars(point.value);
   // The 'Total' cost/point has an 'i' of 0. When queried for the i of its own, non-existent parent, it returns -1.
   const pointIsTotal = point.node.i === 0;
   const parentIsTotal = point.node.parentNode.i === 0;
   const header = pointIsTotal || parentIsTotal ? point.name : [point.node.parentNode.name, point.name].join(": ");
-  return `<b>${header}</b><br/>\n`
-    + `$${abbr.amount} ${abbr.unit}<br/>\n`
-    + `X.Y% of national GDP`;
+  const base = `<b>${header}</b><br/>\n$${abbr.amount} ${abbr.unit}`;
+  return `${base}<br/>\n${((point.value / nationalGdp) * 100).toFixed(1)}% of national GDP`;
```
EmmaLRussell, line 33: Want to include the year here too, or is that obvious enough from the headline figure?

```suggestion
  return `${base}<br/>\n${((point.value / nationalGdp) * 100).toFixed(1)}% of 2018 national GDP`;
```

EmmaLRussell, line 33: This is a tiny bit repeated logic and formatting with the same calculation in cost card. Could go in a shared util - up to you.

EmmaLRussell (commented):
Would  be nice to include in e2e - maybe a regex check rather than a fixed numeric value that would be sensitive to model changes.

components/utils/charts.ts
david-mears-2: Maybe once it's repeated three times!

EmmaLRussell (approved):

PR #63, opened 2024-10-18
Display vaccination time series

david-mears-2:
The changes necessitated by adding a fourth time series.

Restyles time series to allow more vertical space for vaccination time series.


components/TimeSeries.vue
```diff
@@ -84,7 +84,22 @@ const accordionStyle = {
...
+    break;
+  case "dead":
+    yUnits = "deaths";
+    break;
+  case "vaccinated":
+    yUnits = "vaccinated";
+    break;
+  default:
+    yUnits = "cases";
+    break;
```
EmmaLRussell, line 101: It might be safer to just let the unit be empty, though we shouldn't ever really hit the breakpoint.

EmmaLRussell (approved):
Looks great

PR #66, opened 2024-10-18
Use runId from route to load scenario details

david-mears-2:
Now we can use the runId from the route to retrieve the status, results, and parameters for a scenario, rather than relying on pinia to remember the runId from the first page.

This enables users to refresh results pages, and indeed share them.

Some required changes:
* Components that use Highcharts can only be rendered client side (or you'll see errors from Highcharts), so they've been made client-only by renaming them.
* Probably as a result of that, components that use Highcharts now need to explicitly wait for the div that is to contain the chart to be rendered before they can attach the chart. They do this using a ref and a watch, whereas before I could simply initialize the chart in onMounted (but now the div ref is `null` in onMounted).


components/TimeSeriesList.client.vue
```diff
@@ -3,7 +3,6 @@
     <TimeSeries
       v-for="(_, seriesId, index) in appStore.timeSeriesData"
       :key="seriesId"
-      :all-charts="charts"
```
EmmaLRussell, line 6: Why was this not needed any more?

EmmaLRussell (commented):
Great to get this working! Just one tiny question.

pages/scenarios/[runId].vue
```diff
@@ -135,6 +133,13 @@ const paramDisplayText = (param: Parameter) => {
   }
 };
 
+const route = useRoute();
+const runIdFromRoute = route.params.runId as string;
+if (appStore.currentScenario.runId && runIdFromRoute !== appStore.currentScenario.runId) {
+  appStore.clearScenario(); // Required so that previous parameters aren't hanging around in the store.
+}
+appStore.currentScenario.runId = runIdFromRoute;
+
```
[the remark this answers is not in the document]
Comment (david-mears-2) [reply pages/scenarios/[runId].vue:142]: Yes, that's right, otherwise we might be showing parameters from a previous scenario until the new scenario's results come back.

components/TimeSeriesList.client.vue
david-mears-2: I forgot to remove it in a refactor a few weeks ago. It doesn't do anything.

EmmaLRussell (approved):

PR #67, opened 2024-10-21
Jidea-149 show vsl 

absternator:
This PR displays vsl info on results page....

i have just put down generic text to display please review and let me know what to update to.

image


absternator (commented):
Note: revert common api branch after r-api work is merged.. do NOT merge until then

components/CostsCard.vue
```diff
@@ -61,6 +60,9 @@
           @mouseover="hideTooltips = false"
         />
       </div>
+      <p class="text-end fw-lighter vsl-display">
+        Average value of life: {{ formatCurrency(appStore.currentScenario.result.data?.average_vsl || 0) }} USD
+      </p>
```
david-mears-2, line 65: I'm no longer at all sure what the VSL value really represents. @pratikunterwegs , can you review the wording here? We might need to make it clearer that this is something variable / country-dependent, so maybe "National value of life used".

Also, I think I recall that the unit is not USD, but international dollars, but I'm not sure.

david-mears-2 (commented):
I've tagged Pratik for a question.

Now that we have a table, I think a nice tidy place for the information could be in a (?) icon with tooltip, which is the way we present details about time series defintions (already in main) and how we present parameters help text in https://github.com/jameel-institute/daedalus-web-app/pull/57. I'd place this icon with tooltip into the 'Life years' row of the table.

components/CostsCard.vue
pratikunterwegs: I think this is sufficient, and "Value of a statistical life" which is the expansion of VSL would also be appropriate, and perhaps preferable as it suggests that it's a broad generalisation.

It's explicitly not the national value-of-life, but a measure calculated using the country-specific GNI. It generalises well - otherwise, we'd have to implement the specific formula each country uses for its national calculations, including for countries where this isn't strictly defined or readily available.

The unit is indeed international dollars.

pratikunterwegs: On international dollars: https://datahelpdesk.worldbank.org/knowledgebase/articles/114944-what-is-an-international-dollar

david-mears-2:
```suggestion
      <p class="text-end fw-lighter vsl-display">
        Value of a statistical life: {{ formatCurrency(appStore.currentScenario.result.data?.average_vsl || 0) }} Int$
      </p>
```

absternator:
> Now that we have a table, I think a nice tidy place for the information could be in a (?) icon with tooltip, which is the way we present details about time series defintions (already in main) and how we present parameters help text in #57. I'd place this icon with tooltip into the 'Life years' row of the table.

good idea however it refers to pie chart as well... so I've added the * way to show extra info 
image

components/CostsPie.client.vue
```diff
@@ -218,7 +218,7 @@ const populateCostsDataIntoPie = () => {
     costsData.push({
       id: cost.id,
       parent: appStore.totalCost!.id,
-      name: appStore.getCostLabel(cost.id),
+      name: `${appStore.getCostLabel(cost.id)}${cost.id === "life_years" ? "*" : ""}`,
```
david-mears-2, line 221: What does this affect - the label in the tooltips?

components/CostsTable.vue
```diff
@@ -13,7 +13,7 @@
       >
         <tr>
           <td class="ps-2">
-            {{ appStore.getCostLabel(childCost.id) }}
+            {{ appStore.getCostLabel(childCost.id) }} <span v-if="childCost.id === 'life_years'">*</span>
```
david-mears-2, line 16: Think this shouldn't have a space:

```suggestion
            {{ appStore.getCostLabel(childCost.id) }}<span v-if="childCost.id === 'life_years'">*</span>
```

components/CostsPie.client.vue
david-mears-2: I don't think it looks good in the cost pie labels - sorry! I think it's best to just have the * once, in the table.

image

components/CostsCard.vue
```diff
@@ -61,6 +60,9 @@
           @mouseover="hideTooltips = false"
         />
       </div>
+      <p class="fw-lighter vsl-display">
+        * Value of statistical life: ${{ formatCurrency(appStore.currentScenario.result.data?.average_vsl || 0) }} Int'l$
+      </p>
```
EmmaLRussell, line 65: I think this is probably one of those situations where you ought to be able to assume that if we're showing this element then the result data exists. We shouldn't ever be showing 0 here!

EmmaLRussell (approved):
LGTM, one tiny comment relating to our discussion on assumptions and Typescript.

PR #71, opened 2024-10-22
Fix costs pie bug

david-mears-2:
Update chart levels settings when navigating via chart breadcrumbs

Ensures that the central circle (the total) is given non-zero area when navigating from drill-down to top-level by use of the breadcrumb link labelled 'Total'.

The link is shown in this screenshot:

image

Before this change it resulted in:

image

After:

image


EmmaLRussell (approved):
Works for me!

PR #72, opened 2024-10-22
jidea-162 Code snippet

EmmaLRussell:
Adds a button (next to the download button) which opens a dialog showing daedalus code snippet for the current parameters, with link to the daedalus package. Includes button for copying snippet to clipboard.

I've tested that the code snippet runs and generates results, and Pratik has eyeballed it too.

The styling of the Copy button is up for debate - it changes its text (to "Copied!) and background colour on click, then back on mouseleave. I did try tooltip for this but CTooltips don't seem to work well with dynamic content unfortunately (pointer gets stuck in the position for the initial content). 

Because we only need the scenario parameters to generate the snippet, the button is visible and enabled before results are returned. I think that's ok, but maybe it looks a bit odd?


david-mears-2:
[Screencast from 23-10-24 13:54:28.webm](https://github.com/user-attachments/assets/b0377b11-f380-4266-9a85-3a07991d37dc)

Two things - 
1. I think it's suboptimal that the border disappears on hover. I'd prefer just the background color to change, to something other than the page's background color, e.g. a whiter shade.
2. When you hold your mouse down on these buttons, they turn black. This makes a flash if you click slowly, which I didn't quite manage to do in that screen recording. I think I have already modified this away for the parameters button, search for `.parameters-button:active`.

david-mears-2:
Very small quibble on behaviour: can the 'Copied!' version of the button be visible for longer? The way I click, my mouse almost immediately leaves the hitbox, so the text change is just a flash (changes back when not hover).

components/CodeSnippet.vue
```diff
@@ -0,0 +1,107 @@
...
+      </CTooltip>
+    </div>
+    <CModal
+      :visible="modalVisible"
+      aria-labelledby="modalTitle"
+      @close="() => { modalVisible = false }"
+    >
+      <CModalHeader>
+        <CModalTitle id="modalTitle">
+          Daedalus Code Snippet
```
david-mears-2, line 26:
```suggestion
          DAEDALUS code snippet
```
The titlecase looks quite odd.

EmmaLRussell:
> Very small quibble on behaviour: can the 'Copied!' version of the button be visible for longer? The way I click, my mouse almost immediately leaves the hitbox, so the text change is just a flash (changes back when not hover).

Yeah, good idea, I'll put a little timeout on it rather than tie it to mouseleave.

stores/appStore.ts
```diff
@@ -135,7 +135,7 @@ export const useAppStore = defineStore("app", {
         } finally {
           this.downloading = false;
         }
-      })();
+      }, 200)(); // include small delay to avoid tooltip flicker on immediate button disable
```
EmmaLRussell, line 138: This change seemed to be necessitated by the css changes - not really sure how, but I think the issue was with immediately disabling the button while the tooltip was showing - the tooltip was momentarily displayed orphaned from the button on top left of page.

absternator (approved):
works great!! LGTM!

PR #76, opened 2024-10-23
Fix unit tests by updating mocked results response

david-mears-2:
Unit tests were broken on main by https://github.com/jameel-institute/daedalus-web-app/pull/75

I opted to 'fix forward' by updating the mocked results response as well, so that they match. The new mocked results then necessitated various other text expectation changes.


EmmaLRussell:
Is this context of all this the change to using country ISOs, and the front end needing to adapt to that?

EmmaLRussell:
Oh, other changes too, right, new time series etc? So it's not necessarily stuff that the front end needs to know about because it's driven by metadata, but you just wanted to test against a real result set to check everything still works ok?

david-mears-2:
Yes, I hadn't updated the mocked responses for quite some time, so other changes to the API metadata have come along too. This came about because I wanted to be able to access time_series_groups from the mocked metadata.

EmmaLRussell (approved):
It might have made more sense to keep the bits of the mock response that didn't need to change e..g. the existing time series values. But I guess this is probably less fiddly!

PR #78, opened 2024-10-25
Jidea-164 Costs pie use color blind safe colors

absternator:
Have updated costs pie colors to be color blind safe.. used the IBM color blind safe colors... can be seen here at bottom of page --> https://davidmathlogic.com/colorblind/#%23648FFF-%23785EF0-%23DC267F-%23FE6100-%23FFB000

image


EmmaLRussell:
Thanks for this! Chatted it over with Mantra and Pratik, and we reckoned the blue and purple tones were a bit too similar here, for non-colourblind vision - so instead have gone with some colours from a different palette (from the same page you referenced)
image

PR #80, opened 2024-10-28
Fix globe rotation bugs

M-Kusumgar:
This fixes 2 bugs:
1. the globe has a gentle rotation animation however the starting rotation and the ending rotation were not the same points so one the animation stopped (in 180s, you can change it to less time in the code if you want to see it) the globe jumped to the starting rotation, this was just a case of replacing the `-` to a `+` so now the starting and ending rotations have a difference of 360 degs, i.e. no difference visually
2. if you select a country, spin the globe around 2 revolutions or so and pick another country, the globe would reverse your manual rotation and then spin to the country, we would like it to pick the shortest path of rotation so have done that with some maths

### DIFF
--- a/components/Globe.vue
+++ b/components/Globe.vue
@@ -201,10 +201,10 @@ const rotateChart = (direction: "x" | "y", to: number) => {
       diffRotation = diffRotation - 360;
     }
     // gets actual rotation destination by adding the difference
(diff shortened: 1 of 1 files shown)


EmmaLRussell (commented):
LGTM, one tiny suggestion

PR #81, opened 2024-10-28
Add unselectable land color

M-Kusumgar:
Added a slightly faded colour for the unselectable land, felt myself being confused as to which colour meant what and then i thought i could also fade the boundaries slightly of the geojsons and i think this helped. Let me know what you think!

### DIFF
--- a/components/Globe.vue
+++ b/components/Globe.vue
@@ -37,6 +37,7 @@ WHONationalBorders.features = WHONationalBorders.features.map((feature) => {
 const activeCountryColor = am5.color("#379f9f"); // Imperial sea-glass
 const defaultLandColour = am5.color("#002e4e"); // A default land colour: dark blue
 const unselectableLandColor = am5.color("#19425e"); // Color for land with missing data
+const unselectableLandBoundaryColor = am5.color("#19425e"); // Color for stroke of the missing data
 const hoverLandColour = am5.color("#1c6777"); // The midpoint between defaultLandColour and activeCountryColor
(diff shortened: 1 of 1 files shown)


components/Globe.vue
```diff
@@ -83,17 +84,19 @@ const chartDefaultSettings: am5map.IMapChartSettings = {
...
   geoJSON: WHONationalBorders,
-  fill: defaultLandColour,
 };
 // Settings for the 'background' world map, a map that includes all countries regardless
 // of whether they are supported in the model. These are the majority of the world's countries
 // and are not able to be selected.
 const backgroundSeriesSettings: am5map.IMapPolygonSeriesSettings = {
   ...countrySeriesSettings,
+  fill: unselectableLandColor,
+  stroke: defaultLandColour,
```
EmmaLRussell, line 94: so the faded stroke colour you're using here just happens to be the same as the selectable (default land colour)?
Wonder if you should make an alias const to be clear what's going on.  It's also a bit confusing because the boundaries in unselectable regions look a lot lighter than the selectable countries' fill - does the chart take the colour you give it and fade it anyway?

EmmaLRussell (commented):
Looks great, I think!

components/Globe.vue
M-Kusumgar: yh annoyingly the chart auto fades it a bit, idea was to blend the boundaries in as much as possible to i wanted to use the colour the countries are filled in, ill make an alias i think!

PR #82, opened 2024-10-28
jidea-174 hide globe in about page

EmmaLRussell:
Prevents distracting planetary intrusion while perusing the About page.. 
This might not be the most sustainable approach once there are other pages - though I think most of those may have the background globe - but since the Globe is already making decisions based on the route, this seemed a reasonable place to put it.


M-Kusumgar (approved):
awesome, as discussed we should change to a `v-if` after the workshops!

PR #84, opened 2024-10-28
jidea-178 move version info to about page

EmmaLRussell:
Add version info to the bottom of the About page, and remove it as random tooltip on brand logo!

Took this opportunity to update the web app version, which we have not been doing well at!

Arguably we could move the versions fetch out of the default layout and into the About page, but it feels like it makes sense to keep it in the layout, as it is also fetching the metadata. I've added an e2e test to make sure it all works together.

### DIFF
--- a/components/AppHeader.vue
+++ b/components/AppHeader.vue
@@ -19,7 +19,6 @@
           data-testid="ji-logo-header"
           style="width: 150px;"
           class="img-fluid"
-          :title="versionTooltipContent"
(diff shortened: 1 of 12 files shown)


EmmaLRussell:
Amazingly, I've managed to use the wrong branch number for the second time in a day - this branch actually relates to jidea-178: https://mrc-ide.myjetbrains.com/youtrack/issue/JIDEA-178/Move-version-information-into-the-About-page

M-Kusumgar (approved):
Nice one

PR #88, opened 2024-11-04
jidea-191 Add small delay to parameter option tooltips

david-mears-2:
This is to make it slightly less visually chaotic when quickly mousing through different options - avoids obscuring options when you are trying to select them


david-mears-2:
Please review at your leisure!

EmmaLRussell:
Added a ticket for this: https://mrc-ide.myjetbrains.com/youtrack/issue/JIDEA-191/Add-delay-to-parameter-option-tooltips

EmmaLRussell (approved):
It's annoying that the None option is getting placed on the right since the tooltip width was increased. Could slim it down to fit it above again to avoid the obscuring problem? But this is fine too. 

Could you align the package version numbers before merging please?

david-mears-2:
> It's annoying that the None option is getting placed on the right since the tooltip width was increased. Could slim it down to fit it above again to avoid the obscuring problem? But this is fine too.

I agree. But I think making all tooltips narrow enough to prevent this (175px) results in them being too narrow and that's worse overall.

PR #89, opened 2024-11-04
Pass dedupe 'defer' option to useFetch

david-mears-2:
https://nuxt.com/docs/api/composables/use-fetch\#params

The default value for this option is 'cancel' - which means that if a new identical request is made, it cancels any ones that are still waiting for their responses.

On slow networks we saw that about half the time, a status poll request resulted in useFetch cacheing the status as 'running' for a long time (minutes) even though new requests were returning the correct status ('completed'). Given that the dedupe option fixes it, it was presumably because new requests were constantly being cancelled when the polling interval came around again, before they had had a chance to return anything (on a slow network speed).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "daedalus-web-app",
   "type": "module",
-  "version": "0.1.0",
+  "version": "0.1.1",
   "private": true,
   "engines": {
     "node": ">=20"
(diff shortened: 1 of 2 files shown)


david-mears-2:
Local testing:

Before the change, I ran the default parameter set from the homepage 10 times. Of these, 6 took longer than 10 seconds (and probably would have gone on for minutes) while 4 succeeded normally. After the change, I ran the same thing 7 times and each time it succeeded normally.

EmmaLRussell (approved):

PR #92, opened 2024-11-11
JIDEA-181: Extract utility-like functions to separate file from Globe component

david-mears-2:
Pull out helper methods into utils module to make Globe component less monolithic, for readability of Globe component.

Add unit tests for most of these helper methods.


EmmaLRussell (approved):
Thanks for this!

PR #94, opened 2024-11-12
JIDEA-171: Add Antarctica to the globe

david-mears-2:
Has been bugging me since Pratik pointed out it was missing.

This uses the geodata that comes along with amcharts, in the lowest provided resolution.


M-Kusumgar (approved):
sorry for the delay! looks good to me

PR #98, opened 2024-11-22
remove API configure step from working globe branch

EmmaLRussell:
Had issues with Globe component on main, so testing dependencies fix targeting jidea-168, a known working branch.


david-mears-2 (approved):
Seems to work! Let's merge this into the targeted branch so that that branch can pass its tests.

PR #99, opened 2024-12-11
Use data attributes to 'sanity-test' the time series charts

david-mears-2:
Add data attributes to the containers of the time series and of the costs pie chart; put the rendered data in there; spot-check it for time series in e2e tests, and check everything for costs pie (small amount of data).

Using 50% +/- tolerance to accommodate future model changes that alter the precise numbers, this is a 'sense-check' to see if the numbers are in the expected order of magnitude.


nuxt.config.ts
```diff
@@ -61,4 +61,6 @@ export default defineNuxtConfig({
       rApiBase: "", // https://nuxt.com/docs/getting-started/testing#registerendpoint
     },
   },
+
+  compatibilityDate: "2024-12-11",
```
david-mears-2, line 65: I'm mainly adding this in to silence the warning about it that happens on starting up the dev server, on the theory that not doing so causes it to be added dynamically, which possibly triggers a full app re-load on start-up (as updating this config normally does), which would slow down the development experience.

Docs: https://nuxt.com/docs/api/nuxt-config#compatibilitydate

components/CostsPie.client.vue
```diff
@@ -5,6 +5,7 @@
       :id="chartContainerId"
       ref="chartContainer"
       :class="[props.hideTooltips ? hideTooltipsClassName : '']"
+      :data-test="JSON.stringify(costsData)"
```
EmmaLRussell, line 8: Maybe call this something a bit less generic - `data-summary`?

EmmaLRussell (commented):
Works a treat!

I don't really love stringfying JSON into HTML attributes, but I think that just an aesthetic squeamishness on my part. An alternative approach would be the one we took in WODIN where we had a separate summary component which would encode individual summary values into separate attributes. But you would need two of those, one for time series and one for (each) costs value. Probably not worth it. 

Maybe include expected number of points for time series as well as first and last?

EmmaLRussell:
> I suppose there's a chance there are missing data points, but that seems like an unlikely scenario?

Maybe, depends how it's implemented. I couldn't remember if the day (x) value came from the data returned by the backend, or if the front end derives it from the index. Less likely in the latter case! (but perhaps the implementation might change... )

david-mears-2:
> Maybe, depends how it's implemented. I couldn't remember if the day (x) value came from the data returned by the backend, or if the front end derives it from the index. Less likely in the latter case! (but perhaps the implementation might change... )

It's derived by index in the front end. We have to draw the line somewhere with sanity-checking the data, and I think this is a decent place to draw it.

david-mears-2:
> It's derived by index in the front end. We have to draw the line somewhere with sanity-checking the data, and I think this is a decent place to draw it.

I changed my mind and added in a test for the number of data points since it doesn't take long to implement.

EmmaLRussell (approved):
Did you get the e2e tests working locally?

david-mears-2:
> Did you get the e2e tests working locally?

No, unfortunately. It was on webkit and mobile safari browsers.

PR #100, opened 2024-12-12
Upgrading dependencies

david-mears-2:
Upgrade Nuxt minor version; run 'npm audit fix' - This introduced extra linting behaviour which was undesirable (re-sorting imports is too noisy in diffs, and anyway by being alphabetical it disrupts separations between external and internal imports) so the relevant rule has been switched off.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "@coreui/icons": "^3.0.1",
     "@coreui/icons-vue": "^2.1.0",
     "@coreui/vue": "^5.2.0",
-    "@nuxt/eslint": "^0.5.3",
+    "@nuxt/eslint": "^0.7.3",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -40,7 +40,7 @@
     "hex-rgb": "^5.0.0",
     "highcharts": "^11.4.8",
     "lodash.throttle": "^4.1.1",
-    "nuxt": "^3.13.2",
+    "nuxt": "^3.14.1592",
```
david-mears-2, line 43: They're doing a pi joke with their version numbers...

M-Kusumgar (approved):
LGTM!

package.json
M-Kusumgar: not sure how i feel about this one :laughing:

PR #102, opened 2025-01-07
Cache the versions endpoint

david-mears-2:
Previously, I gave up trying to cache endpoints because I couldn't see a way to purge the cache between tests so that the tests didn't alter each other's effects. But then this way occurred to me: refer to an environment variable in the app config to decide whether a route is cached or not.


EmmaLRussell (approved):
Nice - request is shown as cached in browser after first fetch

PR #107, opened 2025-03-21
JIDEA-260: Modal for choosing axis (unordered parameters only)

david-mears-2:
## What's the feature?

Introduces a button 'compare against other scenarios' on the scenario results page, which opens a modal. The modal contains a from that first lets the user choose which parameter will be their axis of comparison. It then allows them to specify which parameter _values_ will be included in that comparison. When this modal form is submitted, the user is taken to a placeholder page for the comparison result, which for now is just a table confirming which scenarios the user wanted to see.

There are two corresponding Youtrack tickets: https://mrc-ide.myjetbrains.com/youtrack/issue/JIDEA-227/Multi-select-for-unordered-params-in-modal and https://mrc-ide.myjetbrains.com/youtrack/issue/JIDEA-260/Modals-first-selection-Choosing-the-axis-of-comparison

## How to approach this monstrous PR

Apologies for the size of this PR. (It may be some comfort that 600 lines are just package-lock.json).

### 1. First try manually testing the feature so you can get a feel for what the feature is.

From a scenario result page, you should be able to choose an axis to compare along, some scenarios to be included (but no fewer than 1 and no more than about 5), submit, and then see a table showing you the scenarios you expected.

Pending https://github.com/jameel-institute/daedalus.api/pull/38, currently the only parameters you'll be able to choose to compare along are those that the metadata thinks are unordered and non-numerical: country, and disease.

### 2. Next, note the following:

I originally followed the implementation pattern that the single-scenarios page uses: that when you submit the form, we store all the parameters in the app store and load the results page (/scenarios/:id) based on the details in the store. However, after thinking about the requirement for comparison pages to be able to be shared by link, and the advantage of semantically meaningfully URLs, I decided to make the URL query parameters the source of truth for the comparison information. That means that, when loading a comparison page, the app store populates itself based on the URL query parameters (see e2e test for an example of this). Using semantic URLs has the desirable consequence that comparisons can be shared by link even before we persist comparison records in the db -- which hopefully we'll get done before November, but it's nice to avoid depending on that.

### 3. Review the central files of interest

(You may wish to review the unit tests alongside each corresponding file - each file mentioned in this section has 1 corresponding unit test file.)

- a. components/CreateComparison.vue - contains the button and its modal
- b. components/ScenarioSelect.vue - a component that is used by CreateComparison, and is pretty much a wrapper for our third-party library 'vue3-select-component'. It uses that library to present a multi-select field for specifying which parameter values to include in the comparison.

Both CreateComparison and ScenarioSelect depend on some shared composables:

- c. composables/useComparisonValidation, which handles the validation logic for the form and calculates the feedback to give the user if they do an invalid submission; and
- d. composables/useScenarioOptions.ts, which, based on the user's chosen axis of comparison, works out which scenario options they should be offered. It also converts options from our internal format (based on the metadata) to the vue3-select-component's option format.

The last files of central importance are

- e. pages/comparison.vue - `watch` statement
- f. stores/appStore.ts - `setComparison` function (refer to types/storeTypes.ts for the store interface)

While the comparison.vue page is, as mentioned, a placeholder page, it demonstrates the intended implementation as noted in step 2 above. The lines of interest are those in the `watch` statement, which sets the app store comparison based on the URL query parameters that were set by the `navigateTo` in CreateComparison.

### 4. Review other changes

- No need to review components/ParameterInfoCard.vue : this looks like a new file in the diff, but it isn't new code, it's just extracting a component out of the overly-large pages/[runId].vue
- scripts/run-dev-dependencies : this change improves quality of life for developers by automatically applying the database migrations after start-up of the dependency services.
- tests/e2e/compareScenarios.spec.ts : this provides e2e tests for the two intended journeys to the comparison page: one, via a single scenario result page and the modal form, and two, by following a direct link with a semantic URL.
- utils/utilities.ts : this adds a deepFreeze function copied from our other repos. This is now the third repo with a deepFreeze util, so perhaps we could consider making it into a separate package. Alongside this utility, review the related changes, which are contained in [this commit](https://github.com/jameel-institute/daedalus-web-app/pull/107/commits/a89ea8374009f5db0f23b6bf381f4d58ad02034e), which includes the use of `structuredClone` to avoid copying by reference, and also includes a regression test.
- If you didn't review the unit tests alongside their corresponding files, please do so now.


components/ParameterInfoCard.vue
```diff
@@ -0,0 +1,78 @@
+<template>
```
david-mears-2, line 1: This file doesn't contain new code (except a little tidying of the paramDisplayText function) - it's just been extracted wholesale from [runId].vue in order to make that component smaller.

pages/comparison.vue
```diff
@@ -0,0 +1,70 @@
+<template>
```
david-mears-2, line 1: This page is, obviously, not the final product, it's just a page to test that we are setting the app store values correctly.

pages/comparison.vue
```diff
@@ -0,0 +1,97 @@
...
+  if (scenario.parameters && axis.value) {
+    return scenario.parameters[axis.value];
+  }
+};
+
+const scenarioIsBaseline = (scenario: Scenario) => {
+  return scenarioAxisValue(scenario) === appStore.currentComparison.baseline;
+};
+
+watch(() => appStore.metadata, (newValue) => {
```
EmmaLRussell, line 66: Is this reason that this needs to be a watch on metadata, rather than just happening in `onMounted` that we need to wait for the metadata to be available, and we assume this will happen after the page component has been loaded?.. oh and it also runs immediately so it's fine if metadata is already there..

stores/appStore.ts
```diff
@@ -125,7 +133,25 @@ export const useAppStore = defineStore("app", {
...
-      this.currentScenario = { ...emptyScenario };
+      this.currentScenario = structuredClone(emptyScenario);
+    },
+    clearComparison() {
+      this.currentComparison = structuredClone(emptyComparison);
+    },
+    setComparison(axis: string, baselineParameters: ParameterSet, selectedScenarioOptions: string[]) {
+      this.clearComparison();
+      this.currentComparison.axis = axis;
+      this.currentComparison.baseline = baselineParameters[axis];
```
EmmaLRussell, line 144: You don't want to just re-use `currentScenario` for the baseline stuff? I suppose that might cause problems later on when you can select a different scenario as your baseline, potentially one that you haven't run yet...?
Ah, ok and you're storing all scenario parameter sets in curentComparison.scenarios, including the baseline, and separately keeping a note of what the comparison axis and comparison baseline are.

tests/unit/stores/appStore.spec.ts
```diff
@@ -227,6 +226,60 @@ describe("app store", () => {
       expect(store.downloading).toBe(false);
     });
 
+    it("can set the current comparison based on the axis, baseline parameter, and selected scenario options", async () => {
+      const store = useAppStore();
+      store.setComparison("vaccine", { country: "USA", hospital_capacity: "54321", vaccine: "high", response: "elimination" }, ["none", "low"]);
+
+      expect(store.currentComparison).toEqual({
```
EmmaLRussell, line 233: Could make this less verbose by comparing each scenario with empty scenario spread with the expected parameter values.

tests/e2e/compareScenarios.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+  await selectParameterOption(page, "country", "United States");
+  await page.click(`div[aria-label="${parameterLabels.vaccine}"] label[for="medium"]`);
+  await page.fill(`input[aria-label="${parameterLabels.hospital_capacity}"][type="number"]`, "305000");
+
+  await page.click('button:has-text("Run")');
+
+  await page.waitForURL(new RegExp(`${baseURL}/${scenarioPathMatcher}`));
+  await expect(page.getByText("Simulate a new scenario")).not.toBeVisible();
+
+  // Run a second analysis with a different parameter, using the parameters form on the results page.
```
EmmaLRussell, line 27: this comment feels like it doesn't belong here...?

tests/e2e/compareScenarios.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+  const scenarioSelect = page.getByLabel(`Compare baseline scenario ${baselinePathogenOption} against`);
+  await expect(scenarioSelect).toBeVisible();
+
+  // No options should be pre-selected
+  await expect(page.getByRole("button", { name: "Covid-19 wild-type" })).toHaveCount(0);
+
+  // Open the select
+  await scenarioSelect.click();
+
+  // await page.getByPlaceholder("Select up to 5 options to compare against baseline").click();
```
EmmaLRussell, line 42: can remove this?

tests/e2e/compareScenarios.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+  await page.getByRole("option", { name: "Covid-19 wild-type" }).click();
+  await page.getByRole("option", { name: "Covid-19 Omicron" }).click();
+
+  // The two options should now be listed in the multi-select
```
EmmaLRussell, line 46: Could check they've been removed from the options list too..?

components/utils/formatters.ts
```diff
@@ -0,0 +1,5 @@
+// Convert strings to human readable format (i.e. with comma-separated thousands).
+// TODO: Localize number formatting.
+export const humanReadableNumber = (num: string): string => {
```
EmmaLRussell, line 3:
```suggestion
export const humanReadableInteger = (num: string): string => {
```

components/CreateComparison.vue
```diff
@@ -0,0 +1,205 @@
...
+  chosenAxisId.value = "";
+};
+
+const handleChooseAxis = (axis: Parameter) => {
+  if (chosenAxisId.value === "") {
+    chosenAxisId.value = axis.id;
+    // Pre-populate the scenario options input
+    selectedScenarioOptions.value = nonBaselineOptions.value.length + 1 > MAX_COMPARISON_SCENARIOS
+      ? [] // TODO: (jidea-230) pre-populate country parameter to nearby countries
+      : nonBaselineOptions.value.map(o => o.id);
```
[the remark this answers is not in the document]
david-mears-2: > I don't really understand why previously selected axis makes a difference?

The two top-level branches here are "user is choosing an axis, so populate the field if applicable" and "user is un-choosing axis, so reset our refs to being empty" (in this case, the field is also re-hidden from view). We tell the difference between choosing and un-choosing based on whether choseAxisId already has a value.

components/CreateComparison.vue
```diff
@@ -0,0 +1,205 @@
...
+}, { deep: 1 });
+
+const submitForm = async () => {
+  if (formInvalid.value) {
+    showFeedback.value = true;
+    return;
+  }
+
+  showFeedback.value = false;
+  appStore.downloadError = undefined;
```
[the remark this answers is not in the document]
david-mears-2: (This is the store property for any errors on trying to download.) This line is the equivalent of the same line in the ParameterForm component, which clears appStore.downloadError on submitting the form. I think we did that because otherwise the download error message continues to be displayed if you go back home and then run a new scenario. [Here](https://github.com/jameel-institute/daedalus-web-app/pull/41#discussion_r1796731577) is the thread where we came up with this solution. But probably we can come up with a single place we can reset it rather than resetting it in multiple places - a good candidate for this would be to do it inside [appStore.clearScenario](https://github.com/jameel-institute/daedalus-web-app/blob/7d5dec289e5318b862ac484be7a4d99e6a43623b/stores/appStore.ts#L127)

components/ScenarioSelect.vue
```diff
@@ -0,0 +1,117 @@
...
+      input-id="scenarioOptions"
+      :is-menu-open="menuOpen"
+      :aria="{ labelledby: labelId, required: true }"
+      class="form-control"
+      :class="showFeedback ? 'is-invalid' : ''"
+      :options="nonBaselineSelectOptions"
+      :is-clearable="true"
+      :is-multi="true"
+      :close-on-select="false"
+      :placeholder="`Select up to ${MAX_COMPARISON_SCENARIOS - 1} options to compare against baseline`"
```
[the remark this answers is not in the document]
david-mears-2: Yes - MAX... is the maximum number of scenarios that can be included in a comparison. Here we already know one of the scenarios -- the baseline -- and are asking for the remainder.

components/ScenarioSelect.vue
```diff
@@ -0,0 +1,117 @@
...
+      // Only do anything if the click was on the parent element
+      if (event.target === vueSelectControl.value) {
+        // If a click is detected in a row-end gap that is created due to flex-wrapping the option tags onto multiple rows,
+        // treat it as a click on the search input, that is, open the options menu and focus the search input.
+        menuOpen.value = true;
+        vueSelectControl.value?.querySelector<HTMLInputElement>(SEARCH_INPUT_SELECTOR)?.focus();
+      }
+    });
+  }, { immediate: true });
+});
```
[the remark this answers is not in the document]
david-mears-2: Actually, the library does implement 'click on the select area to open it' in general, and this is a small patch.

When there are enough options selected that the tags wrap onto multiple rows, only clicks on the search-input field, which lies to the right of the options on the last row, are registered. But, if the wrapping happens to work out such that there is a big row-end gap on a non-final row, it feels natural that clicking in that gap should still open the select.

components/ScenarioSelect.vue
```diff
@@ -0,0 +1,117 @@
...
+    }
+  }
+
+  ::placeholder {
+    font-size: var(--cui-body-font-size);
+  }
+
+  &.form-control.is-invalid {
+    // Calculate how much the menu needs to be moved down to accommodate the invalid tooltip:
+    // = tooltip font size + (2 * tooltip vertical padding) + tooltip margin + form-control bottom padding + arbitrary extra padding
```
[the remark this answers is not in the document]
david-mears-2: This is a comment, not commented-out code. It explains the ensuing calculation so it can be re-written if any of those values change.

components/utils/parameters.ts
```diff
@@ -0,0 +1,18 @@
+import type { ParameterOption } from "~/types/parameterTypes";
+import type { Option } from "vue3-select-component";
+
+export interface ParameterSelectOption extends Option<string> {
+  value: string
+  label: string
+  description: string
+}
```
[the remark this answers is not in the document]
david-mears-2: That's right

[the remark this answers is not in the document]
david-mears-2: Specifically it wants 'value' where our metadata uses 'id'

pages/comparison.vue
david-mears-2: I don't remember why I chose watch instead of onMounted, but this will all be refactored some day.

stores/appStore.ts
david-mears-2: Yeah. I think I was loath to have any crucial details about the comparison living outside of the currentComparison property, and/or I didn't want to give currentScenario two slightly different responsibilities. This feels like a way to reduce chances of currentScenario and currentComparison getting tangled up.

david-mears-2:
> I can see the benefit of semantically meaningful urls, but this approach is going to mean that we're a bit inconsistent between id-based links and urls with embedded data, since single-scenario links are id-based currently.

Fair point, we should eventually make them consistent either way.

> I feel like once we get to the stage of having actually run all the comparison scenarios then we're going to want to easily pull all of the result sets to reload a comparison, probably by linking all the individual scenario ids with a "comparison id".

Possibly that's how they'll be linked together. In a world before comparison records (which we plan to introduce but who knows when we will), the alternative would be to derive which scenarios are needed from the comparison's settings. We currently store scenarios in the database with an index of hashed parameter values so that they can be looked up by parameters. That will allow for the possibility of using the knowledge of the axis and baseline to get scenario records from the db without need for a comparison record.

> We could potentially support both approaches, one id-based where we can immediately load results, and one semantic, where the back end would attempt to find and load pre-run scenarios matching the requested parameter values - for both comparisons and single scenarios (and could run those scenarios automatically if they don't exist yet). Or maybe shift entirely to semantic for both?

I'm not sure which I prefer, out of id-based or semantic, as the primary approach. Semantic is simpler for right now. Id-based may be more intuitive when we are doing history/bookmarking features.

tests/unit/components/CreateComparison.spec.ts
```diff
@@ -0,0 +1,237 @@
...
+
+    expect(axisOptionsEl.findAll("button")).toHaveLength(1);
+    expect(countryButton.classes()).toContain("bg-primary");
+
+    // Reveals the scenario selection section
+    expect(modalEl.text()).toMatch(/Compare baseline scenario United Kingdom against:/);
+    expect(getComboboxEl(wrapper).exists()).toBe(true);
+    // No country options are pre-selected
+    expect(wrapper.find("button.multi-value").exists()).toBe(false);
+    // TODO: Test tooltip
```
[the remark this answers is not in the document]
david-mears-2: Ah yes. I couldn't find a way to trigger the tooltip being displayed. I think we could just trust the library to do that properly though.

tests/e2e/compareScenarios.spec.ts
david-mears-2: Oops

components/ScenarioSelect.vue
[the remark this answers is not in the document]
david-mears-2: Updated these to refer to the number of non-baseline scenarios per comparison.

david-mears-2:
> I worry that the "Clear" button is so close to the up-down chevron that user might accidentally clear selections when attempting to close the drop down.. though I don't really have any better ideas! Maybe Clear isn't really needed?

I've got rid of it. It would only really be useful if you had dozens of things in there, which you shouldn't.

> Because clicking on the "Select.." area opened the drop down, I was expecting clicking on it again to close it.

This is a little fiddly to implement - ran up against a bug where it immediately closes after opening. I think the current behaviour is OK.

> Could close the drop-down when all options have been selected.

Done.

david-mears-2:
> I wonder if the pattern of wanting to both deep freeze an object, and provide a structured clone method for whenever you want a deep copy could be provided by a util which does both, to save a bit of repetition.

I don't understand how that would save repetition. Each existing call to deepFreeze would still exist, and so would each call to the structuredClone / cloner.

PR #108, opened 2025-04-01
Update the unit test for About page

david-mears-2:
I moved some mocks out of mockPinia.ts into the test file itself, because most of the consts created in that file are for being fed into the `createTestingPinia` at the end of the file.

If you don't like these mocks to live in the test file, I will create a new file for them to live in under `mocks/`.

While tidying, I removed an unused property for 'intervention level'.

### DIFF
--- a/types/resultTypes.ts
+++ b/types/resultTypes.ts
@@ -9,14 +9,8 @@ export interface ScenarioCapacity {
   value: number
 }
 
-export enum InterventionLevel {
-  Light = "light",
-  Heavy = "heavy",
-}
-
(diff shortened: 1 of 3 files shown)


types/resultTypes.ts
```diff
@@ -9,14 +9,8 @@ export interface ScenarioCapacity {
   value: number
 }
 
-export enum InterventionLevel {
```
absternator, line 12: how come this is r

absternator (approved):
Allgood to me, as long as you don't think the mocks you created will be used anywhere else

PR #110, opened 2025-04-02
Highlight the 'edit parameters' button when the baseline value's tooltip is shown

david-mears-2:
This is an iteration on top of https://github.com/jameel-institute/daedalus-web-app/pull/107

There is a tooltip introduced by that PR that is shown when the user hovers over the baseline value inside the 'create comparison' modal. The tooltip reads "To change the baseline, exit the overlay and edit the current scenario's parameters."

The intention here is to make it clearer to the user where the relevant button is. I reuse the pulse animation we already have for drawing attention to form fields, and apply it to the button, but alongside an '.infinite' class so that we can precisely control the pulse switching on and off.

There are three interactions that can affect the pulse: hovering over the tooltip (starts pulse); closing the 'create comparison' modal (stops pulse after 2 seconds); and exiting the tooltip (stops pulse after 3 seconds).

The pulse durations differ because I was trying to strike a balance between (1) allowing the pulse to continue long enough after closing/exiting that the user sees it and (2) stopping it before it becomes annoying. The pulse is less obvious when the modal is visible, hence the longer duration on exiting the tooltip.

Because there are two triggers for stopping the pulse, each with their own different timeout, and we want to prioritise the effect of the most recent interaction (the most recent timeout to be created), all stop-triggers are routed through a single function, `turnOffEditParamsButtonPulse`, which replaces any already-existing timeout with the new timeout.

### DIFF
--- a/components/CreateComparison.vue
+++ b/components/CreateComparison.vue
@@ -50,6 +50,8 @@
             <CTooltip
               content="To change the baseline, exit the overlay and edit the current scenario's parameters."
               placement="top"
+              @show="() => { $emit('toggleEditParamsButtonPulse', true) }"
+              @hide="turnOffEditParamsButtonPulse(3000);"
             >
               <template #toggler="{ togglerId, on }">
                 <!-- TODO: use humanReadableNumber formatter for numeric parameters -->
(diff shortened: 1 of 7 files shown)


components/CreateComparison.vue
```diff
@@ -93,27 +95,44 @@
...
+}>();
+
 const appStore = useAppStore();
 const FORM_LABEL_ID = "scenarioOptions";
 const selectedScenarioOptions = ref<string[]>([]);
 const modalVisible = ref(false);
 const chosenAxisId = ref("");
 const formSubmitting = ref(false);
 const showFeedback = ref(false);
+const pulseOffTimer = ref<ReturnType<typeof setTimeout>>();
```
M-Kusumgar, line 113: i believe this is just `NodeJS.Timeout` type, but this also works!

M-Kusumgar (approved):
Looks good to me, but on a side note to this PR perhaps:

Why cant the user just change their baseline in the create comparison modal? and we just automatically rerun the model with existing parameter values + their new baseline for whatever field they select and compare that with the 5 other comparison options theyve selected? the dark background of the model definitely made it quite hard for me to see the pulsing, like if i didnt know about it, idk if i wouldve spotted it

david-mears-2:
> the dark background of the model definitely made it quite hard for me to see the pulsing, like if i didnt know about it, idk if i wouldve spotted it

I think this is the strongest argument for discarding this PR.

david-mears-2:
I'm going to mark this PR as draft as it's not a priority feature.

PR #114, opened 2025-05-20
Replace costs pie chart with a column chart, having a cost-basis toggle (%GDP/USD)

david-mears-2:
Replaces the pie chart with a bar chart that displays the same information (well, except that the chart now has no representation of the grand total cost).

The 'cost basis' is whether the costs are shown in 'absolute' terms of USD (as outputted from the model) or are displayed in terms of the percent of national GDP. A toggle controls this setting.

Some other implementation-detail changes are:
* I re-implemented 'resetting everything (tooltips and crosshairs) when the mouse leaves a chart', in order to do so using Highcharts' API rather than the status quo of setting props (`hideTooltips`) that add/remove classes which target certain highcharts elements. This is explained at length in the commit message of https://github.com/jameel-institute/daedalus-web-app/pull/114/commits/6d86a299f69f892e2b61d5ad86fe29d80e7f6827.
 
This PR also contains updates to the costs table:
* Can be expanded/collapsed
* The cost basis is controlled by the global cost basis toggle

When testing locally, you might have to npm install, because I upgraded the highcharts version.

### DIFF
--- a/components/CostsChart.client.vue
+++ b/components/CostsChart.client.vue
@@ -218,17 +218,5 @@ onUnmounted(() => {
 #costsChartContainer {
   font-weight: normal !important;
   z-index: 10; // Above timeseries
-
-  .highcharts-tooltip {
-    transition: filter 0.2s;
-  }
-
-  .highcharts-point {
-    opacity: 1 !important; // Avoid transparency on hover over pie slices
-  }
-
-  .highcharts-container {
-    margin-left: auto;
-  }
 }
 </style>


components/CostsChart.client.vue
```diff
@@ -0,0 +1,233 @@
...
+#costsChartContainer {
+  font-weight: normal !important;
+  z-index: 10; // Above timeseries
+
+  .highcharts-tooltip {
+    transition: filter 0.2s;
+  }
+
+  .highcharts-point {
+    opacity: 1 !important; // Avoid transparency on hover over pie slices
```
david-mears-2, line 226: check if this is needed still

PR #119, opened 2025-06-17
JIDEA-234: Replace costs pie chart with a column chart, having a cost-basis toggle (%GDP/USD)

david-mears-2:
Replaces the pie chart with a bar chart that displays the same information (well, except that the chart now has no representation of the grand total cost).

The 'cost basis' is whether the costs are shown in 'absolute' terms of USD (as outputted from the model) or are displayed in terms of the percent of national GDP. A toggle controls this setting.

If you notice issues with tooltips, please be patient for the ensuing PR that deals with that: https://github.com/jameel-institute/daedalus-web-app/pull/121
 
Updates to the costs _table_ (e.g. toggling the cost basis) will be handled in a follow-up PR: https://github.com/jameel-institute/daedalus-web-app/pull/120/

When testing locally, you might have to npm install, because I upgraded the highcharts version.

### DIFF
--- a/tests/e2e/runAnalysis.spec.ts
+++ b/tests/e2e/runAnalysis.spec.ts
@@ -191,6 +191,9 @@ test("Can request a scenario analysis run", async ({ page, baseURL }) => {
   checkBarChartDataIsDifferent(costsChartDataUsd, costsChart2Data);
   checkBarChartDataIsDifferent(costsChartDataGdp, costsChart2Data);
 
+  // Test that the second analysis retains the user's preference for the "as % of 2018 GDP" cost basis.
+  await expect(page.getByLabel("as % of 2018 GDP")).toBeChecked();
+
   // Test that the user can navigate to previously-run analyses, including when the page is initially rendered server-side.
   await page.goto(urlOfFirstAnalysis);
   await page.waitForURL(urlOfFirstAnalysis);


components/utils/highCharts.ts
```diff
@@ -10,13 +13,33 @@ export const plotLinesColor = "#FF0000"; // red;
...
+export const colorBlindSafeColors = [
   { name: "Vermillion", hex: "#d55e00" },
   { name: "Bluish green", hex: "#009e73" },
   { name: "Sky blue", hex: "#56b4e9" },
 ];
 
-export const costsPieColors = ["rgba(1,1,1,0)"].concat(Array.from<string>(colorBlindSafeColors.map(color => color.hex))); // Make the center circle translucent
+// Create a range of N colors varying in brightness
+// NB This won't work well with colors that are already very light or very dark.
+export const getColorVariants = (colorHex: string, numberOfVariants: number) => {
```
EmmaLRussell, line 24: Is this mimicking what highcharts as doing for us anyway with the colour variants in the pie chart? It doesn't do something similar for stacked bars?

components/CostsChart.client.vue
```diff
@@ -0,0 +1,222 @@
...
+const seriesData = ref<Highcharts.SeriesColumnOptions[]>([]); // This only exists for testing purposes
+const chartContainer = ref<HTMLElement | null>(null);
+const chartParentEl = computed(() => chartContainer.value?.parentElement);
+
+// A range of colors per column, to be used for the breakdowns within each column.
+const columnColors = computed((): string[][] => {
+  return appStore.totalCost?.children?.map((cost, i) => {
+    const columnBaseColor = colorBlindSafeColors[i].hex;
+    const numberOfColorVariants = Math.max(cost.children?.length || 1);
+    return getColorVariants(columnBaseColor, numberOfColorVariants);
```
EmmaLRussell, line 36: Minor performance saving, but you could just calculate these colour variants once in the util, since they're always used in conjunction with the colour blindness safe colours, rather than tying it to the component load.

components/utils/highCharts.ts
```diff
@@ -10,13 +13,33 @@ export const plotLinesColor = "#FF0000"; // red;
...
+  const maxBrightness = 1.25; // 125% brightness
+  for (let i = 0; i < numberOfVariants; i++) {
+    const factor = i / (numberOfVariants - 1);
+    const brightness = minBrightness + (maxBrightness - minBrightness) * factor;
+    let r = Math.round(rgb.red * brightness);
+    let g = Math.round(rgb.green * brightness);
+    let b = Math.round(rgb.blue * brightness);
+    r = Math.max(0, Math.min(255, r));
+    g = Math.max(0, Math.min(255, g));
+    b = Math.max(0, Math.min(255, b));
```
EmmaLRussell, line 37: This is potentially going to change the hue of the variants isn't it? e.g. if you start off with an r, g or b value which gets clipped at its max while the others don't?
Would it make more sense to work in hsl? Maybe using [this](https://www.npmjs.com/package/color-convert) if we haven't already got a converter which will do it.

tests/unit/components/CostsCard.spec.ts
```diff
@@ -58,23 +56,21 @@ describe("costs card", () => {
...
     const component = await mountSuspended(CostsCard, { global: { stubs, plugins } });
 
-    const costsPieComponent = component.findComponent({ name: "CostsPie.client" });
-    expect(costsPieComponent.props().hideTooltips).toBe(false);
+    const costsChartComponent = component.findComponent({ name: "CostsChart.client" });
+    expect(costsChartComponent.props("basis")).toBe(CostBasis.USD);
 
-    const container = component.find(`#costsPieContainer`);
+    // const costsTableComponent = component.findComponent({ name: "CostsTable" });
+    // expect(costsTableComponent.props("basis")).toBe(CostBasis.USD);
```
EmmaLRussell, line 66: Can this commented out code be removed or will it be reinstated?

tests/unit/components/CostsCard.spec.ts
```diff
@@ -58,23 +56,21 @@ describe("costs card", () => {
...
-    await container.trigger("mouseleave");
-    await waitFor(() => {
-      expect(costsPieComponent.props().hideTooltips).toBe(true);
-    }, { timeout: 1000 });
+    await nextTick();
 
-    await container.trigger("mouseover");
-    expect(costsPieComponent.props().hideTooltips).toBe(false);
+    expect(costsChartComponent.props("basis")).toBe(CostBasis.PercentGDP);
+    // expect(costsTableComponent.props("basis")).toBe(CostBasis.PercentGDP);
```
EmmaLRussell, line 74: ..same here

tests/unit/components/CostsChart.spec.ts
```diff
@@ -0,0 +1,337 @@
...
+      expect.objectContaining({ y: 0, name: "" }),
+      expect.objectContaining({
+        name: "Retirement-age adults",
+        y: 175405.3478,
+      }),
+    ],
+  }),
+];
+
+describe("costs pie", () => {
```
EmmaLRussell, line 153:
```suggestion
describe("costs chart", () => {
```

david-mears-2:
> Did we talk about potentially showing the USD cost as the primary figure when you haven't got "Show losses as % of GDP" switched on i.e. should be larger and on the left? The view is a little bit confusing at the moment, defaulting to USD as primary but the big figure you see first is the total GDP % value.

Yes. I've listed that ticket as a 'Could' for now: https://mrc-ide.myjetbrains.com/youtrack/issue/JIDEA-283/When-cost-basis-is-toggled-update-the-cost-card-text-to-foreground-the-relevant-cost

david-mears-2:
> Do we have any plans to save user's toggle preferences? I can imagine it getting a bit annoying if you prefer your figures in GDP but they re-default to USD on each run.

Good point.

components/CostsChart.client.vue
```diff
@@ -0,0 +1,222 @@
...
+        },
+      },
+    },
+    series: getSeries(),
+    legend: {
+      enabled: false,
+    },
+    tooltip: {
+      shared: true,
+      distance: 128, // necessary for shared tooltips to not obscure stack labels or grand-child costs within a stack.
```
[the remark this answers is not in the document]
david-mears-2: I think I was trying to prevent overlaps like this:

image

However, we now get overlaps like this instead:

Screenshot from 2025-06-25 11-23-03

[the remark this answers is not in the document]
david-mears-2: Do people 'bite' dilemmas? I was thinking of bullets.

components/utils/highCharts.ts
david-mears-2: Yeah, there's nothing equivalent for stacked bars. Highcharts provides `colorByPoint`, but this produces variation _within_ a series, which is an entity that is orthogonal to the stacked bar.

components/utils/highCharts.ts
david-mears-2: What a good idea

david-mears-2: I used your hsl idea, and here it is working:
image

david-mears-2:
> The curse of Firefox again (squashed blobs - not an issue on Chrome). I don't think I've got Firefox set up weirdly (e.g. Zoom is 100%)..

I'm electing not to investigate this...

david-mears-2:
> I'm assuming the over-sticky tooltips are addressed in the next PR..

A veritable assumption!

components/CostsChart.client.vue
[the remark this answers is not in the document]
EmmaLRussell: I associate horns more with drinking than biting..

components/utils/highCharts.ts
EmmaLRussell: Ha, these variants seem a bit blander, I think I preferred them before! :facepalm: Up to you!

EmmaLRussell (approved):
Looks great to me!

Could potentially add an e2e test that the costs basis preference is persisted if you wanted to.

david-mears-2:
> Could potentially add an e2e test that the costs basis preference is persisted if you wanted to.

This needs only a single line in the e2e so I'm adding it in.

PR #120, opened 2025-06-17
JIDEA-284: Update the costs table to have a configurable cost basis and be collapsible

david-mears-2:
Contains these updates to the costs table:

* Can be expanded/collapsed
* The cost basis is controlled by the global cost basis toggle

### DIFF
--- a/components/CostsCard.vue
+++ b/components/CostsCard.vue
@@ -73,7 +73,7 @@
       <div class="chart-and-table-container">
         <CostsChart id="costsChartContainer" />
         <div class="flex-grow-1">
-          <CostsTable data-testid="costs-table" :basis="costBasis" />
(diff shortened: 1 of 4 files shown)


absternator (approved):
Nice work !! works well and code lgtm!

components/CostsCard.vue
```diff
@@ -73,7 +73,7 @@
       <div class="chart-and-table-container">
         <CostsChart id="costsChartContainer" />
         <div class="flex-grow-1">
-          <CostsTable data-testid="costs-table" />
+          <CostsTable data-testid="costs-table" :basis="costBasis" />
```
david-mears-2, line 76: This prop should not be here - left over from previous version

PR #121, opened 2025-06-18
Refactor the synching and hiding of tooltips and crosshairs

david-mears-2:
'Synchronising' tooltips and crosshairs is the behaviour you can see among the 4 time series, whereby hovering at time point N on one chart creates tooltips on all 4 charts at the same time point (and same for crosshairs).

In the past, following the lead of [a Highcharts demo](https://www.highcharts.com/demo/highcharts/synchronized-charts), we replaced the `reset` function of the Highcharts.Pointer prototype with 'undefined', which appeared to be required for complicated reasons (if you're interested in those reasons, my hypothesis is it seemed to be required because the demo's implementation calls `onMouseOver` to highlight synchronised charts' points, and that function implicitly calls `reset`).

However, this had the undesirable effect of preventing unrelated charts (such as our costs bar chart, formerly a pie chart) behaving normally, since the prototype is shared by all Highcharts charts. As I recall, the issue was that (in the costs pie) the tooltips failed to get reset on the mouse leaving the container; thus we had to either manually reset this unrelated chart, or manually apply classes to hide the tooltips on mouse leave (we had gone with that latter option, using a `hideTooltips` prop and class to target certain highcharts elements). Lots of overhead, just because an unrelated chart had mucked about with Highcharts internals.

I found a better way of doing things, which uses Highcharts' API. I wanted a solution that would allow other charts to behave as expected by default - then people can implement new charts without having to know all this context about why their tooltips aren't working properly.

My solution is for the time series (synchronised charts) to use Highcharts' 'wrap' function (for extending/hooking into private functions) to wrap the Pointer.onContainerMouseLeave function. In this wrapper, we check if the chart that is triggering onContainerMouseLeave is marked as a synchronised chart (and is in the relevant group of charts), in which case we should prevent the wrapped function from executing; otherwise, it must be another unrelated chart, and so we allow the original function to proceed. A second patch-up is then necessitated for synchronised charts, since onContainerMouseLeave is suppressed for those charts, and that is what had otherwise been responsible for triggering Pointer.reset() - thus, we must manually trigger the Pointer.reset() for all synch'd charts when the mouse leaves the collection.


EmmaLRussell (commented):
Thanks for the detailed explanation! Works great. 

I feel like this approach is a little unvue-y in the way it has the top-level component get a list of charts from Highcharts and manipulate those directly, circumventing the chart components which each know about their own charts. 

Would an alternative approach be for the top level Card component to set a "hoverPoint" prop in the child TimeSeries components (via the mid-level group component I guess) which the TimeSeries can watch and respond to as appropriate. It splits up the logic but it does mean you don't need to keep the indices record and filter the full list of charts in the Card, which is additional potentially fragile logic.

There may be subtle reasons I'm not aware of why this wouldn't work though!

PR #122, opened 2025-06-18
Sort options in comparison modal scenario selection, if parameter is deemed 'ordered' by the metadata

david-mears-2:
Whether a parameter is 'ordered' is already configured in the metadata provided by the R API. E.g. the "none", "low", "medium" and "high" options for global vaccine investment clearly have an ordering to them. But the list of diseases doesn't have an ordering.

The idea is that this configuration should fix the order that options are displayed in the scenario select, both in terms of the 'tags'/'pills' within the control, and the options in the dropdown menu.

Relates to https://github.com/jameel-institute/daedalus.api/pull/38 in that that PR will update the API's metadata to make the 'response' parameter unordered.

### DIFF
--- a/components/utils/parameters.ts
+++ b/components/utils/parameters.ts
@@ -1,4 +1,4 @@
-import type { ParameterOption } from "~/types/parameterTypes";
+import type { Parameter, ParameterOption } from "~/types/parameterTypes";
 import type { Option } from "vue3-select-component";
 
 export interface ParameterSelectOption extends Option<string> {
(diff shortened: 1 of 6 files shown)


components/utils/parameters.ts
```diff
@@ -16,3 +16,25 @@ export const paramOptsToSelectOpts = (options: Array<ParameterOption>): Paramete
...
+// If the parameter is ordered, sort the options based on their order in the metadata.
+export const sortOptions = (parameter: Parameter, optionsToSort: string[]) => {
+  if (parameter.parameterType === "numeric") {
+    return optionsToSort.sort((a, b) => {
+      return Number.parseInt(a) - Number.parseInt(b);
+    });
+  }
+
+  if (!parameter.ordered || parameter.options === undefined) {
+    // Special case for unordered parameters where a 'none' option exists: such parameters can be considered semi-ordered.
```
M-Kusumgar, line 33:
```suggestion
    // Special case for unordered parameters where a 'none' option exists: such parameters can be considered semi-ordered and 'none' will be placed at the start.
```

M-Kusumgar (approved):
LGTM, had a tiny comment improvement, would be useful at a glance to know where "none" goes in the sort order!

PR #124, opened 2025-06-24
JIDEA-280: Vary hospital capacity among comparison scenarios, depending on country

david-mears-2:
While defining the parameter values that we want all scenarios in a comparison to share, we need to replace any numeric parameters that are dependent on other parameters with their default values, so that the comparison scenarios are commensurable. For example, when comparing a scenario in Switzerland with one in the USA, we need the hospital capacities to differ, because the countries have very different populations.

I also added a new getter to appStore for easy parameter metadata lookup.


david-mears-2:
Hang on a sec, this doesn't work in the slightest

PR #125, opened 2025-06-24
Use flags in country selects and replace flag icon source

david-mears-2:
## Use flags in country selects

image

image

## Replace flag icon source

image

CoreUI flag icon libraries are updated very rarely, and have a few errors like the flags for South Africa and South Korea being way too small.

The new flag icon library has the desirable feature of normalizing flag shapes to a 4x3 ratio.


M-Kusumgar (approved):
Looks good to me, just to confirm if flag icons doesnt know the particular iso2 code, it is just a blank space right?

david-mears-2:
> Looks good to me, just to confirm if flag icons doesnt know the particular iso2 code, it is just a blank space right?

That's right.

PR #129, opened 2025-07-04
JIDEA-280: Vary hospital capacity among comparison scenarios, depending on country

david-mears-2:
Normally, a comparison is a set of scenarios that differs only on one parameter 'axis', e.g. low/medium/high vaccine investment.

But sometimes we want one axis to vary in sympathy with another. For example, when comparing a scenario in Switzerland with one in the USA, we need the hospital capacities to differ, because the countries have very different populations. We say that hospital_capacity is 'dependent' on country, and this is configured in the parameter metadata.

Therefore, while defining the parameter values for a set of scenarios in a comparison, we need to replace any dependent numeric parameters with their default values, so that the comparison scenarios are commensurable.

The 'create comparison' modal warns the user that they are about to have two parameters varying, if applicable:

image

and if the baseline differs from the default value, we also describe what behaviour to expect there:

image

I also added a new getter to appStore for easier parameter metadata lookup.

## Manual testing

Please check the parameters of the scenarios created when you create a comparison.

### DIFF
--- a/components/CreateComparison.vue
+++ b/components/CreateComparison.vue
@@ -45,6 +45,29 @@
           </CButton>
         </div>
         <div v-if="chosenParameterAxis && baselineOption" class="mt-3">
+          <div v-if="paramsDependingOnAxis?.length" class="alert alert-warning">
+            <p>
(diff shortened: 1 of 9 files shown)


david-mears-2:
Thanks for the design suggestion, I've updated so it will come out like this:

image

PR #132, opened 2025-07-23
Add multi-scenario costs chart, with toggle-able cost basis

david-mears-2:


components/utils/highCharts.ts
```diff
@@ -152,21 +158,64 @@ export const costsChartTooltipText = (context: unknown, costBasis: CostBasis, na
...
+// Tooltip text for a stacked column in a multi-scenario costs chart (shared tooltip for all points in the stack)
+export const costsChartMultiScenarioStackedTooltip = (context: unknown, costBasis: CostBasis, axisParam: Parameter | undefined) => {
+  const contextInstance = context as TooltipPointInstance;
+  const point = contextInstance.point;
+  if (!point || !axisParam) {
+    return;
+  }
+
+  const scenarioCategory = point.category as string;
+  const scenarioLabel = axisParam.options?.find(o => o.id === scenarioCategory)?.label || humanReadableInteger(scenarioCategory);
```
EmmaLRussell, line 170: Need to allow for the integer option here because of hospital capacity?

components/utils/highCharts.ts
```diff
@@ -152,21 +158,64 @@ export const costsChartTooltipText = (context: unknown, costBasis: CostBasis, na
...
+
+  let headerText = `${axisParam.label}: <b>${scenarioLabel}</b>`;
+
+  if (costBasis === CostBasis.PercentGDP) {
+    const percentOfGdp = humanReadablePercentOfGdp(point.total);
+    headerText = `${headerText}</br></br>Total losses: <b>${percentOfGdp.percent}%</b> ${percentOfGdp.reference}`;
+  } else {
+    const abbreviatedTotal = abbreviateMillionsDollars(point.total);
+    const totalCostAsGdpPercent = point.points?.map(p => p.custom.costAsGdpPercent).reduce((sum, a) => sum + a, 0);
```
EmmaLRussell, line 179: Why do you need to calculate this from the individual points rather than from the total?
Could you also move this calculation inside the conditional branch and only calculate if point.total > 0?

components/utils/highCharts.ts
```diff
@@ -152,21 +158,64 @@ export const costsChartTooltipText = (context: unknown, costBasis: CostBasis, na
...
+export const costsChartYAxisTickFormatter = (value: string | number, costBasis: CostBasis) => {
   if (costBasis === CostBasis.PercentGDP) {
     return `${value}%`;
   } else if (costBasis === CostBasis.USD) {
     return Object.values(expressMillionsDollarsAsBillions(value as number, 0, true)).join(" ");
   }
 };
+
+export const costsChartMultiScenarioXAxisLabelFormatter = (value: string, axisParam: Parameter | undefined) => {
+  const text = axisParam?.options?.find(o => o.id === value)?.label || humanReadableInteger(value);
```
EmmaLRussell, line 211: This is a little bit of shared code with costsChartMultiScenarioStackedTooltip - could pull out into something like `getValueLabel(value, axisParam)`

stores/appStore.ts
```diff
@@ -86,6 +92,23 @@ export const useAppStore = defineStore("app", {
...
+      return state.currentComparison.scenarios?.map(s => s.result.data?.costs).every(c => !!c && c.length > 0);
+    },
+    axisLabel(state): string | undefined {
+      if (state.currentComparison.axis) {
+        return this.parametersMetadataById[state.currentComparison.axis].label;
+      }
+    },
+    baselineScenario(state): Scenario | undefined {
+      return state.currentComparison.scenarios.find((scenario) => {
+        if (state.currentComparison.axis && scenario.parameters) {
```
EmmaLRussell, line 105: I guess you could return undefined from the top level if !state.currentComparison.axis, rather than getting into the find at all. 

Why would scenario.parameters be falsy?

EmmaLRussell (commented):
Looks great! The flags are a really nice touch. :)

I can't remember what we decided about indicating baseline scenario in the comparison charts. As far as I can tell, the only indicator is that it's the first bar depicted. I think It would be nice to have a little more - could maybe put the bar label in bold if that's easy to do, or in a different colour and/or with (Baseline) label like you've done in the table. 

The tooltips are still hanging around after the cursor leaves the chart (on Firefox) - is that expected?

components/CompareCostsChart.client.vue
```diff
@@ -0,0 +1,152 @@
...
+const scenarios = computed(() => appStore.currentComparison.scenarios);
+const costBasis = computed(() => appStore.preferences.costBasis);
+const chartTitleOnExport = computed(() => `Losses by ${appStore.axisLabel?.toLocaleLowerCase()}`);
+const axisMetadata = computed(() => appStore.currentComparison.axis ? appStore.parametersMetadataById[appStore.currentComparison.axis] : undefined);
+
+// There are 3 levels of data breakdown for costs:
+// 1) the top-level total for the scenario,
+// 2) the second level, GDP/life-years/education,
+// 3) and a further breakdown, e.g. absences/closures in the cases of GDP and education.
+// We don't display the top level in the chart.
```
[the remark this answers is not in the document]
david-mears-2: Anyway, good point, and I'll update the comment

components/CompareCostsChart.client.vue
```diff
@@ -0,0 +1,152 @@
...
+// Series are orthogonal to columns, and here correspond to the second-level breakdowns.
+// Each series' data is an array of each scenario's cost for that breakdown.
+// For example, one series should comprise each scenario's GDP.
+const getSeries = (): Highcharts.SeriesColumnOptions[] => {
+  const exampleSecondLevelCosts = scenarios.value[0].result.data?.costs[0].children;
```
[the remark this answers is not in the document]
david-mears-2: That's right. Commented.
`// Take the first scenario's costs as an example to find out what the second-level breakdowns are.`

components/CompareCostsChart.client.vue
```diff
@@ -0,0 +1,152 @@
...
+  } as Highcharts.SeriesColumnOptions)) || [];
+
+  return seriesData.value;
+};
+
+const chartHeightPx = 400;
+const yAxisTitle = computed(() => costBasis.value === CostBasis.PercentGDP
+  ? `Losses as % of ${gdpReferenceYear} national GDP`
+  : "Losses in billions USD",
+);
```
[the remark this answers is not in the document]
david-mears-2: yAxisTitle can indeed be shared.

I think the chart height is subject to change as the app develops, so I'll keep it as a separate const here.

components/CompareCostsChart.client.vue
```diff
@@ -0,0 +1,152 @@
...
+
+// Fixes z-index issue:
+// https://www.highcharts.com/docs/chart-concepts/labels-and-string-formatting#html
+Highcharts.HTMLElement.useForeignObject = true;
+
+const chartInitialOptions = () => {
```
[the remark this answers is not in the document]
david-mears-2: I've already tried to extract every last thing in this options object that's shared into the highCharts.ts utils file. So anything left in either chart component is something specific to that component. It still takes up 50-odd lines in this component, and 80 in the other.

We could still move this into a separate file and set the options conditionally based on a boolean 'single' vs 'multi' parameter - but I think that would be to the detriment of keeping related code in one place.

components/utils/highCharts.ts
david-mears-2: Moving the calculation inside results in some fairly nested conditionals that I'm not sure how to improve upon:

```ts
if (costBasis === CostBasis.PercentGDP) {
    const percentOfGdp = humanReadablePercentOfGdp(point.total);
    headerText = `${headerText}</br></br>Total losses: <b>${percentOfGdp.percent}%</b> ${percentOfGdp.reference}`;
  } else {
    const abbreviatedTotal = abbreviateMillionsDollars(point.total);
    headerText = `${headerText}<br/></br>Total losses: <b>$${abbreviatedTotal.amount} ${abbreviatedTotal.unit}</b>`;
    if (point.total > 0) {
      const totalCostAsGdpPercent = point.points?.map(p => p.custom.costAsGdpPercent).reduce((sum, a) => sum + a, 0);
      if (totalCostAsGdpPercent) {
        const percentOfGdp = humanReadablePercentOfGdp(totalCostAsGdpPercent);
        headerText = `${headerText}</br>(${percentOfGdp.percent}% ${percentOfGdp.reference})`;
      }
    }
  }
```

components/utils/highCharts.ts
david-mears-2: This is now shared as `getScenarioCategoryLabel`.

stores/appStore.ts
david-mears-2: > I guess you could return undefined from the top level if !state.currentComparison.axis, rather than getting into the find at all.

True, but the TS linter objects to that, for some reason (`Type 'undefined' cannot be used as an index type`), whereas it doesn't when the check is closer to the usage (`scenario.parameters[state.currentComparison.axis]`).

> Why would scenario.parameters be falsy?

The type signature for the property is `ParameterSet | undefined`. A specific situation where it might be undefined is in the last few lines of appStore.loadScenarioDetails : if the request for the scenario fails, we don't assign any parameters to the scenario.

EmmaLRussell (approved):
Looks great, thanks!

PR #133, opened 2025-07-23
Update parameter info card

david-mears-2:
This PR encompasses some styling changes for the heading row of two pages, the single-scenario results page and the multi-scenario results page, and makes some visual updates to the parameter info card, which were requested by the EPPI folks. They wanted it to be more obvious that the EditParameters button (now labelled 'Change parameters', formerly labelled 'Parameters') was a button one can interact with. To that end, I've updated its name, position in the card, and background color. I also removed the different layout (header vs footer) that depended on the page width, as I found it had become too complicated (judging the maintenance costs as too high).

In order to use the Parameter Info Card in the multi-scenario page, I refactored it so that the content of the header is now a slot. This lets us put an EditParameters button in the header for the single-scenario page, and a link to the baseline scenario for the multi-scenario page.


pages/comparison.vue
```diff
@@ -205,6 +217,17 @@ onUnmounted(() => {
...
 
+.results-cards-container {
+  // The widest width that might trigger the heading flex row to wrap
+  // (Determined empirically using longest heading 'Explore by global vaccine investment' and longest parameter values)
+  $comparison-heading-wrap-width: 91rem;
+
+  @media (min-width: $comparison-heading-wrap-width) {
+    position: relative;
+    top: -2.5rem;
+  }
```
david-mears-2, line 228: The purpose of this is to bump up the tabs when there is space available to do so, saving vertical space.

PR #136, opened 2025-08-04
Highlight the baseline scenario in column chart

david-mears-2:
The baseline scenario will be denoted / highlighted by:
- Different border color for column (white vs transparent)
- Different opacity (baseline scenario is 100% opaque)
- Noted in the x-axis label as '(baseline)'


david-mears-2:
Feedback from EPPI meeting: varying the color using opacity (by too much) is not a good idea if multiple colors here have the same hue (i.e. we use multiple blues). It makes them become too similar. So either reduce/stop using opacity for highlighting, use differenter colours for this chart, or both.

david-mears-2:
font-weight

PR #138, opened 2025-08-07
Jidea 258 time series comparison

david-mears-2:


david-mears-2:
Feedback from EPPI meeting: showing interventions vertically stacked like screenshot is confusing, because it looks like the y-axis might mean something. A good suggestion was to reveal the scenario-specific intervention period only when hovering over the relevant scenario (default to displaying the baseline's intervention period).

image

david-mears-2:
A todo: show the hospital capacity line as dotted, so it's easier to distinguish from the timeseries lines (in a separate PR, given this will require changes to the ChartLegend component). https://mrc-ide.myjetbrains.com/youtrack/issue/JIDEA-294/Show-capacity-line-as-dotted-so-its-easier-to-distinguish-from-the-timeseries-lines

components/utils/highCharts.ts
```diff
@@ -169,7 +196,7 @@ export const costsChartSingleScenarioTooltip = (context: unknown, costBasis: Cos
 };
 
 // 'Category' is the internal name for the x-axis in a multi-scenario costs chart, i.e. it denotes which scenario the column refers to.
-const getScenarioCategoryLabel = (category: string, axisParam: Parameter | undefined): string => {
+export const getScenarioCategoryLabel = (category: string, axisParam: Parameter | undefined): string => {
```
david-mears-2, line 199: todo, move to components/utils/comparisons.ts

david-mears-2: and rename to getScenarioLabel

PR #145, opened 2025-08-27
Add a 'signpost' for advanced users to discover the model package

david-mears-2:
Because some external users were conflating the dashboard and the model, I thought it would be good to add some kind of hint / signpost to make it more obvious that the model does a lot more than the dashboard does, allowing:

(1) non-technical users to understand that the model and the dashboard are different products, where the former allows more control, and the latter is merely a 'window into' the possibilities afforded by the model; and

(2) more advanced users (R users) to discover the model package.

We already hint at the existence of the model / separation of the model and dashboard in (1) the sidebar, which has a link labelled 'DAEDALUS GitHub', and (2) the About page, but these things are not themselves very salient.

I thought a good place to leave a hint to this effect could be in the 'edit parameters' modal because it's where people might look for more granularity over parameters. I also added it to the 'create comparison' modal (displayed once a parameter axis is selected). Here is what the hint looks like: it's a button styled to look like a link labelled 'Advanced usage'. Clicking the link on 'here' closes the current modal and opens the code snippet modal.

Relevant component nesting in case helpful:

```
[runId].vue
├── CodeSnippet
├── CreateComparison
│   └── AdvancedUsagePopover
└── EditParameters
    └── ParameterForm
        └── AdvancedUsagePopover
```

## edit params modal

Screenshot from 2025-08-27 16-33-27

Screenshot from 2025-08-27 16-33-23

## create comparison modal

Screenshot from 2025-08-28 15-17-59

Screenshot from 2025-08-28 15-18-06


pratikunterwegs:
I think this is useful and good, happy for it to be included in this way.

absternator (approved):
Nice work looks great!

PR #147, opened 2025-09-02
Small refactorings of time series components

david-mears-2:
A "time series group" is a pair of related time series, like 'hospital demand' and 'new hospitalisations per day', that can be switched between using a toggle. When the time series is changed, we keep the same chart and container going, and just update the data and options of the chart.

The changes:

1. Rename type interface TimeSeriesGroup to TimeSeriesGrouping to avoid collision with component of the same name.

1. We're needlessly changing the id of the container elements for time series charts when they are toggled to the other time series in a group. This could potentially confuse Highcharts. So I made the container element names constant by basing them on the series group index instead of the series itself.


david-mears-2:
Close in favour of https://github.com/jameel-institute/daedalus-web-app/pull/154

PR #148, opened 2025-09-04
Bugfix for Globe

david-mears-2:
Occasionally, due to a missing `selectableCountriesSeries`, the Globe would throw an error, which would prevent a page navigation from happening. I'm not sure why it goes missing - presumably there is some race condition - but this makes the app more usable in that we don't want to prevent navigations just because the globe (which is decorative) has something wrong with it.


absternator (approved):
nice fix

PR #149, opened 2025-09-05
Update cost basis units' presentation

david-mears-2:
We were surfacing the info that GDP data comes from 2018, but we were doing so in the wrong place - it should be when we're displaying data in terms of USD, not in terms of % of GDP, since to interpret the USD figure you need to know how big the economy is assumed to be (and adjust upwards, probably), whereas GDP is naturally proportional to itself (a 20% hit to the economy is always going to be the same percent size no matter the absolute size of the economy).

So "2018" is most relevant as an explanation when talking in dollar terms.

## USD units referring to 2018:

table:
Screenshot from 2025-09-05 11-16-29

cost basis toggler tooltip:
Screenshot from 2025-09-05 11-16-05

costs column chart, y axis and tooltip:
Screenshot from 2025-09-05 11-15-53

### DIFF
--- a/tests/e2e/runAnalysis.spec.ts
+++ b/tests/e2e/runAnalysis.spec.ts
@@ -176,7 +176,7 @@ test("Can request a scenario analysis run", async ({ page, baseURL }) => {
   });
 
   // Check that after toggling the cost basis we see different data.
-  await page.getByLabel("as % of GDP").check();
+  await page.getByLabel("as % of pre-pandemic GDP").check();
   const costsChartDataGdpStr = await page.locator("#costsChartContainer").getAttribute("data-summary");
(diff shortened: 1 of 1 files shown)


pratikunterwegs:
Thanks, I think this makes sense. One other reason to specify the year is exchange rate fluctuations, but that's for user analysts to explain or process downstream.

david-mears-2:
Rob said that this way of expressing units was misleading - "in 2018 USD" sounds like it means the figures are in terms of dollars that have the value dollars did in 2018. Whereas what we really mean to communicate is that our data on how big the economy was at the start of the pandemic comes from 2018, and users may wish to account for the economy being differently sized now.

PR #153, opened 2025-09-16
Adapt VSL help texts to emphasise subjectivity and adjustableness

david-mears-2:
The help text has been moved from a tooltip to a modal because we needed clickable links.

Recent EPPI discussion touched on this change. The intention behind it is to make the following fact more salient: that this VSL number is a subjective-ish choice, and one that can be adjusted by the user in R (at the moment you have to hover to even find out that there exists a VSL value). Crucially we are using the word  ‘assumed’. We also now link to the methodology and usage guide (vignette).

image

image

image

image


components/CostsTable.vue
```diff
@@ -102,10 +87,60 @@
       </template>
     </tbody>
   </table>
+  <p class="fw-lighter small mt-auto">
+    *Based on an assumed value of statistical life:
+    click<CIconSvg class="icon help-icon filter-to-primary p-0 ms-1" role="button">
```
EmmaLRussell, line 92: What's CIconSvg giving us here when we're just going to put an image tag inside it anyway rather than any svg? I think you could get the same behaviour by just putting the class and role attributes on  - that seems to be how it gets rendered anyway.

EmmaLRussell (commented):
This looks good! I don't think it's completely obvious how to actually change the VSL value in R from the linked vignette - I guess you could just edit it in the country object? But that's not a problem with this UI!

components/CostsTable.vue
david-mears-2: You make a good point. I think the only benefit of this is to clue in the developer that "icon" is a CSS class originating from CoreUI. Replacing with img!

PR #156, opened 2025-09-26
Jidea-295 Comparison download

EmmaLRussell:


stores/appStore.ts
```diff
@@ -273,12 +274,16 @@ export const useAppStore = defineStore("app", {
         await this.loadScenarioResult(scenario);
       }) || []);
     },
-    async downloadExcel() {
+    async downloadExcel(comparison: boolean) {
```
david-mears-2, line 277: We're missing some updates to the appstore unit tests: these still say `store.downloadExcel()` whereas they should be taking an argument.

david-mears-2:
It also occurs to me that it's a little odd to have a top-level folder `download`, implying an equal status to things as broad as `tests`, `assets`, and `docker`, to name a few. Would you be opposed to nesting `download` under the top-level folder `utils`?

EmmaLRussell:
> It also occurs to me that it's a little odd to have a top-level folder `download`, implying an equal status to things as broad as `tests`, `assets`, and `docker`, to name a few. Would you be opposed to nesting `download` under the top-level folder `utils`?

I'm not a huge fan of stuffing disparate things into `utils` tbh - my preference would be to have a top level `src` folder, to distinguish app source code from tests,docker, assets etc, under which would go `components`, `layouts`, `composables` etc. I feel like `download` is a sizable and self-contained enough feature to merit its own folder under `src` in that case. That would certainly be a refactor for another ticket though!

PR #172, opened 2025-11-13
Add some error handling, feeding back errors to the user

david-mears-2:
We currently feed back only a small set of the kinds of errors that could happen.

I thought we had better cover the possibility of the Nuxt API returning non-200 responses (which it might do if something gets overloaded by having many requests to deal with / runs hogging all the cores), so that users don't just watch a spinner going round forever.

The kinds of errors that are fed back to the user, as of this PR, are now:

* When a request to _run_ a scenario gives a non-200 status (which could originate with the R API or with our app)
  * for single-scenario runs (from the parameter form)
  * for multi-scenario runs (from the 'create comparison' form)
* When a _status_ request for some scenario gives a non-200 status (which could originate with the R API or with our app)
  * for single-scenario results page
  * for multi-scenario results page
* When a _result_ request for some scenario gives a non-200 status (which could originate with the R API or with our app)
  * for single-scenario results page
  * for multi-scenario results page
* When the R API reports that a model run was unsuccessful, forward this through to the front end.
  * for single-scenario runs (this was already being handled)
  * for multi-scenario runs

## manual testing

To simulate the fetch errors, add a line such as `throw new Error` in the relevant files such as `server/api/scenarios.post.ts` and `server/api/scenarios/[runId]/result.get.ts` and `server/api/scenarios/[runId]/status.get.ts`. Or in `server/api/handlers/scenarios.ts` if you prefer.

## unit tests

For each of the functions 'refreshScenarioStatus' and 'loadScenarioResult', the [runId] page component may call them in two different places: during script setup (just in case data is already available), and after the page is mounted. Thus, when unit-testing the [runId] page, I sometimes tested a couple of possibilities, e.g.: `"shows alert when a status request throws an error during page setup script"` and `"shows alert when a status request throws an error during polling"` since I wanted to cover both code paths.


components/ParameterForm/ParameterForm.vue
```diff
@@ -8,6 +8,21 @@
       novalidate
       @submit.prevent="submitForm"
     >
+      <template v-if="submitError">
+        <CAlert color="danger">
+          <CAlertHeading>
+            <CIcon icon="cilWarning" class="flex-shrink-0 me-2" width="24" height="24" />
+            Error
+          </CAlertHeading>
```
EmmaLRussell: There's an argument to made for pulling out the common error display stuff into a component, but I'd be happy for that refactor to be done after the workshop if you make a ticket for it!

EmmaLRussell (commented):
This looks great for the workshop, couple of suggestions for possible refactors later.

I guess there's also the possibility of not being able to get an api response to load metadata on initial app load. But I found this difficult to sabotage! Even if I killed the api container, i was still seeing parameter options being loaded to the front end. Are they cached in the Nuxt backend?

Are you sure you want to take out the jobSlow feedback? This could be very useful feedback if we really are expecting some slow responses during times of heavy use. Could tweak the threshold times and messages to current requirements.

pages/scenarios/[runId].vue
```diff
@@ -27,33 +27,33 @@
...
-      </p>
-      <p>
-        Analysis status: {{ appStore.currentScenario.status.data?.runStatus }}
-      </p>
-      <p class="mb-0">
-        Waiting for {{ secondsSinceFirstStatusPoll }} seconds
-      </p>
-    </CAlert>
+    <CSpinner v-if="showSpinner" class="ms-3 mb-3 mt-3" />
+    <template v-if="scenarioStatusResponseError || scenarioResultResponseError || appStore.currentScenario.status.data?.runSuccess === false">
```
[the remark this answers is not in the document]
Comment (david-mears-2) [reply pages/scenarios/[runId].vue]: Now I have
```
There was an unexpected error.
          {{ appStore.currentScenario.status.data?.runSuccess !== false ? "Please refresh the page." : "Please try again later." }}
```

components/ParameterForm/ParameterForm.vue
david-mears-2: jidea-319 now exists

components/ParameterForm/ParameterForm.vue
```diff
@@ -212,7 +229,13 @@ const submitForm = async () => {
...
 
   appStore.currentScenario.parameters = { ...formData.value };
-  await appStore.runScenario(appStore.currentScenario);
+  try {
+    await appStore.runScenario(appStore.currentScenario);
+  } catch (error) {
+    submitError.value = error as NuxtError<{ message?: string }>;
+    formSubmitting.value = false;
+    return;
+  }
```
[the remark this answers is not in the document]
david-mears-2: I've made the implementations more similar (they differ slightly due to some requests calling `$fetch` and others `useFetch`) by centralising errors into the store (they all live on a scenario object now).

david-mears-2:
To simulate a slow response, edit the function rApiScenarioStatusEndpoint to this:

```ts
const rApiScenarioStatusEndpoint = (runId: string) => `/scenario/status/${runId}`;
export const getScenarioStatus = async (
  runId: string | undefined,
  event?: H3Event<EventHandlerRequest>,
): Promise<ScenarioStatusResponse> => {
  if (!runId) {
    const errors: Array<ApiError> = [{ error: "Bad request", detail: "Run ID not provided." }];
    return badRequestResponse(errors) as ScenarioStatusResponse;
  }

  const response = await fetchRApi<ScenarioStatusData>(
    rApiScenarioStatusEndpoint(runId),
    {
      method: "GET",
    },
    event,
  );

  response.data = {
    runStatus: "running" as runStatus,
    runSuccess: null,
    done: false,
    runErrors: null,
    runId: response.data?.runId ?? null,
  };

  return apiResponse<ScenarioStatusData>(response) as ScenarioStatusResponse;
};
```

pages/scenarios/[runId].vue
```diff
@@ -27,33 +27,43 @@
         </template>
       </ParameterInfoCard>
     </div>
+    <CAlert v-if="showJobSlowAlert" color="info">
+      <p>
+        Thank you for waiting. At times of heavy usage, scenarios can sometimes take up to 10 seconds to run.
```
Comment (EmmaLRussell) [inline pages/scenarios/[runId].vue]: Up to you, but you might want to be more vague than this! 😆

EmmaLRussell (commented):
This isn't quite working how I'd expect when there's an error submitting a scenario, having previously successfully run a scenario. Instead of staying on the Parameters form, it briefly flashes up the error then goes to the results page with error message to refresh the page. In this case it seems to be using an old scenario id - if I go back and try to run again, it does the same thing, and navigates to the results page for the same scenario id. I guess the run id should be cleared when we hit the submit button - keeping the previous run ID at that point will just confuse things. 

(It's navigating because of this logic in ParameterForm:
```
 if (appStore.currentScenario.runId) {
    await navigateTo(`/scenarios/${appStore.currentScenario.runId}`);
  }
```
This isn't a new bug since the last review, I think I just hadn't tested for it before.

EmmaLRussell:
> I'm struggling to reproduce this... was it from the parameter form in the modal, or the front page? When exactly was the dummy error being thrown?

Try this:
- Run up the app and deps locally
- Run a scenario as usual - note the run id in the address of the Results page
- Navigate back to Parameters Form
- Kill all running containers so the API is gone
- Change some parameter values
- Click Run
- See that error briefly is shown in form, but then it navigates to results page. No results can be loaded, but note that it's still the previous run id in address

david-mears-2:
Reminder to self to increment version number

david-mears-2:
> > I'm struggling to reproduce this... was it from the parameter form in the modal, or the front page? When exactly was the dummy error being thrown?
> 
> Try this:
> 
> * Run up the app and deps locally
> * Run a scenario as usual - note the run id in the address of the Results page
> * Navigate back to Parameters Form
> * Kill all running containers so the API is gone
> * Change some parameter values
> * Click Run
> * See that error briefly is shown in form, but then it navigates to results page. No results can be loaded, but note that it's still the previous run id in address

I never saw the flashing but I have reproduced the mistaken navigation, so I've put it a check for 'run' fetch errors.

EmmaLRussell (approved):
Great, navigation issue now fixed. I do think it would be worth revisiting some of the conditional logic for navigation, button disabling etc, as it is getting quite complex, and may be simplifiable by strategically clearing some store values - but that's definitely a post-workshop refactor!

PR #175, opened 2025-11-17
jidea-318 update about page

EmmaLRussell:
According to updated content posted in Teams chat. 

I have included the derived count of pathogens, however this perhaps is less useful as there is now a hardcoded table listing R0 and fatality ratio for each pathogen .


EmmaLRussell:
I've added a non-clever hardcoded table of contents..

david-mears-2 (approved):

PR #181, opened 2026-04-28
Don't close time series accordion when tapping tooltip icon on touchscreen device

david-mears-2:
E.g. 
image

You can compare this branch on your touchscreen device against prod. Please ignore e2e tests for now, these are pending updates due to model changes - will handle in separate PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented unintended actions when interacting with the tooltip trigger by stopping event propagation and switching to a dedicated button trigger.

* **Accessibility**
  * Added an explicit accessible label and improved hit-area and touch behavior for the tooltip trigger.

* **Tests**
  * Updated tooltip tests to target the new button trigger and added a test ensuring icon interactions don't emit unrelated events.

* **Chores**
  * Bumped package version to 1.0.2.

### DIFF
--- a/components/TooltipHelp.vue
+++ b/components/TooltipHelp.vue
@@ -26,6 +26,9 @@
         <img
           :src="`/icons/${infoIcon ? 'info.png' : 'circleQuestion.svg'}`"
           :aria-describedby="togglerId"
+          @click.stop
+          @pointerdown.stop
+          @touchstart.stop
           v-on="on"
         >
       </CIconSvg>
--- a/components/TooltipHelp.vue
+++ b/components/TooltipHelp.vue
@@ -19,19 +19,21 @@
       </template>
     </template>
     <template #toggler="{ togglerId, on }">
-      <CIconSvg
-        class="icon help-icon opacity-50 p-0"
+      <button
+        type="button"
+        class="tooltip-help-trigger"
         :class="classes"
+        :aria-describedby="togglerId"
+        aria-label="Show help information"
+        v-on="on"
+        @click.stop
+        @pointerdown.stop
+        @touchstart.stop
       >
-        <img
-          :src="`/icons/${infoIcon ? 'info.png' : 'circleQuestion.svg'}`"
-          :aria-describedby="togglerId"
-          @click.stop
-          @pointerdown.stop
-          @touchstart.stop
-          v-on="on"
-        >
-      </CIconSvg>
+        <CIconSvg class="icon help-icon opacity-50 p-0">
+          <img :src="`/icons/${infoIcon ? 'info.png' : 'circleQuestion.svg'}`">
+        </CIconSvg>
+      </button>
     </template>
   </CTooltip>
 </template>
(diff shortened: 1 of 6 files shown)


david-mears-2:
> 🧹 Nitpick comments (1)
> > tests/unit/components/Charts/TimeSeriesTile.spec.ts (1)> `149-151`: **Cover touchscreen event paths in this regression test.**
> > > The fix includes `pointerdown`/`touchstart` propagation guards, but this test only checks `click`. Adding those events here will better protect the actual touchscreen bug.
> > > 
> > > 🧪 Suggested test adjustment
> > > ```diff
> > > -    await component.find("img").trigger("click");
> > > +    const tooltipIcon = component.find(".help-icon img");
> > > +
> > > +    await tooltipIcon.trigger("click");
> > > +    await tooltipIcon.trigger("pointerdown");
> > > +    await tooltipIcon.trigger("touchstart");
> > > 
> > >      expect(component.emitted("toggleOpen")).toBeUndefined();
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against the current code and only fix it if needed.
> > > 
> > > In `@tests/unit/components/Charts/TimeSeriesTile.spec.ts` around lines 149 - 151,
> > > The test currently only triggers a "click" on the image; update the spec around
> > > the image interaction (the component.find("img") test) to also trigger
> > > "pointerdown" and "touchstart" events before the click to cover
> > > touchscreen/propgation-guard paths introduced by the fix, e.g. call
> > > component.find("img").trigger("pointerdown") and
> > > component.find("img").trigger("touchstart") (and then .trigger("click")) and
> > > keep the assertion that component.emitted("toggleOpen") is undefined to ensure
> > > touch and pointer events are handled the same as mouse clicks.
> > > ```
> 
> 🤖 Prompt for all review comments with AI agents
> ℹ️ Review info

Good comment, will implement

M-Kusumgar (approved):
This works well

curious as to why add a button around it with that class instead of just putting
```
@click.stop
@pointerdown.stop
@touchstart.stop
```
on `CIconSvg` and calling it a day, is it because of accessibility? either way works and looks good to go

david-mears-2:
> This works well
> 
> curious as to why add a button around it with that class instead of just putting
> 
> ```
> @click.stop
> @pointerdown.stop
> @touchstart.stop
> ```
> 
> on `CIconSvg` and calling it a day, is it because of accessibility? either way works and looks good to go

It was to increase the hitbox - my finger couldn't reliably find the button otherwise!

david-mears-2:
> > This works well
> > curious as to why add a button around it with that class instead of just putting
> > ```
> > @click.stop
> > @pointerdown.stop
> > @touchstart.stop
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > on `CIconSvg` and calling it a day, is it because of accessibility? either way works and looks good to go
> 
> It was to increase the hitbox - my finger couldn't reliably find the button otherwise!

Probably also more in line with accessibility if clickable things are buttons

PR #186, opened 2026-05-02
Reset timers when navigating on the client side

david-mears-2:
Contrary to the code comment being removed, the most common type of navigation is actually client-side navigation (e.g. when you change parameters from a given scenario to another). In such cases we need to reset the record of the time of the first status poll, so that alert messages telling the user 'you've been waiting X seconds' actually reset back to 0 after a navigation. (This also applies to navigating from a scenario to a comparison, and then via a baseline to a new comparison, when all those navigations are client-side i.e. via clicking UI buttons instead of typing in the address bar - and it's easier to see the bug in this case because comparisons tend to take longer than single scenarios, and so you're more likely to see the 'it's taking a while' alert when doing a comparison.)

The bug was that this number was not being reset for client-side navigation, so each time you saw an alert message about a long wait period, it was telling you the seconds since you requested some earlier thing, not the current thing you are waiting for.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of job timing detection on comparison and scenario pages by refining how initial status timestamps are captured and calculated.

### DIFF
--- a/pages/comparison.vue
+++ b/pages/comparison.vue
@@ -104,6 +104,7 @@ const query = useRoute().query;
 
 const jobSlow = ref(false);
 const secondsSinceFirstStatusPoll = ref("0");
+const timeOfFirstStatusPoll = ref(new Date().getTime());
 
 const showSpinner = computed(() => {
   return !everyScenarioIsDone.value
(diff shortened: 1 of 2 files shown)


david-mears-2:
@coderabbitai Before I believe the analysis from the review agents I'd like you to reproduce the bug they think they've identified: run up the app in your container and verify/falsify that the bug happens

david-mears-2:
@coderabbitai Please do so, since their claim does not seem to be borne out for me locally. Are they talking about navigating via typing in the address bar as opposed to client-side navigations?

david-mears-2:
@coderabbitai I was asking you to verify the comments from the agents, not the bugfix of the PR (which was mine)

david-mears-2:
Coderabbit has falsified one of its issues; and the second issue, the query-based navigation issue, is pre-existing. It's also not a live issue because we don't actually do any navigation via changing a query parameter.

absternator (approved):
works as intended and code looks lgtm!

PR #192, opened 2026-05-19
trigger ci workflows

david-mears-2:
## Summary by CodeRabbit

* **Improvements**
  * R API requests are now staggered to distribute traffic evenly over time, improving system stability.

* **Tests**
  * Added verification that queued R API requests respect stagger intervals.

[Review Change Stack](https://app.coderabbit.ai/change-stack/jameel-institute/daedalus-web-app/pull/192?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/server/utils/rApi.ts
+++ b/server/utils/rApi.ts
@@ -18,6 +18,7 @@ let nextRApiRequestStart = 0;
 
 const wait = (ms: number) => new Promise<void>(resolve => setTimeout(resolve, ms));
 
+// Test-only helper; do not call during request handling.
 export const resetRApiRequestStagger = () => {
   nextRApiRequestStart = 0;
 };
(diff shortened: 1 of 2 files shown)


david-mears-2:
@copilot+claude-opus-4.7 Pass a critical eye over the changes in this PR. Would the solution implemented address the problem they are intended to address, namely taking load off the R API server when front-end comparison requests trigger multiple quasi-simultaneous requests to the R API server, without thereby increasing load on the web app server? And separately, how would the effects of the changes differ if all the requested scenarios had already been run and thus their results were cached/pre-canned in the web app db - would the web app still make requests to the R API server or does it totally route around the R API server in that situation?

david-mears-2:
@copilot+claude-opus-4.7 Given we can assume that most requested scenarios have been run already (pre-canned), I want to temporarily put in place server-side cacheing of result requests (to be backed out after today's launch). What is the simplest way to do this? is it updating routeRules in nuxt.config to be

```ts
  routeRules: {
      "/": { redirect: "/scenarios/new" },
      "/api/versions": cacheRoute(60),
      "/api/metadata": cacheRoute(60),
      "/api/scenarios/*/result": cacheRoute(86400),
    },
```

or does that only cache per-user?

david-mears-2:
> 2\. `maxAge: 86400` will also cache error responses by default

This is an unacceptable risk. How do we prevent this in Nuxt? @copilot

PR #193, opened 2026-05-26
De-dupe scenario select options

david-mears-2:
'Scenario-selecting' refers to choosing the scenarios you want to include in a comparison. The multi-select input that initiates a comparison pre-loads some suggestions/defaults. For the case of comparing scenarios by hospital capacity, it pre-loads 3 suggested values: the minimum capacity, default capacity, and maximum capacity that are configured for the current country. But that implementation assumes those 3 values are unique, which is not always the case. For very small countries, the minimum and/or maximum may be the same as the default (see e.g. [Costa Rica](https://github.com/jameel-institute/daedalus-web-app/blob/176a31e2f99b3a915cda8c3166fd1718a009983c/mocks/responses/metadata.json#L482)). In such cases we should de-duplicate the suggested/pre-loaded options, so e.g. when you are in Costa Rica and set hospital capacity to 700, the options presented (when you attempt to compare by hospital capacity) should be 600 and 800, rather than 600, 600 and 800:

Screenshot from 2026-05-26 11-10-26

And post the fix:

image

So this branch addresses the edge-case where some scenario select options are the same, by de-duplicating them.

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved duplicate entries in predefined numeric options and ensured the current baseline option is excluded so scenario option lists show unique, relevant values.

* **Tests**
  * Added unit coverage verifying deduplication of predefined numeric options and the baseline exclusion behaviour.

[Review Change Stack](https://app.coderabbit.ai/change-stack/jameel-institute/daedalus-web-app/pull/193?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/composables/useScenarioOptions.ts
+++ b/composables/useScenarioOptions.ts
@@ -50,7 +50,8 @@ export default (parameterAxis: MaybeRefOrGetter<Parameter | undefined>) => {
       return [];
     }
     if (axis.value?.parameterType === TypeOfParameter.Numeric) {
-      return predefinedNumericOptions.value?.filter(o => o.id !== baselineOption.value?.id) || [];
+      const opts = predefinedNumericOptions.value?.filter(o => o.id !== baselineOption.value?.id) || [];
+      return opts.filter((opt, index, self) => index === self.findIndex(o => o.id === opt.id)); // dedupe by id
     }
     return axis.value?.options?.filter(({ id }) => {
       return baselineOption.value && id !== baselineOption.value?.id;
--- a/composables/useScenarioOptions.ts
+++ b/composables/useScenarioOptions.ts
@@ -1,6 +1,7 @@
 import { type Parameter, type ParameterOption, TypeOfParameter } from "~/types/parameterTypes";
 import { getRangeForDependentParam, paramOptsToSelectOpts } from "~/components/utils/parameters";
 import { commaSeparatedNumber } from "~/components/utils/formatters";
+import type { DisplayInfo } from "~/types/apiResponseTypes";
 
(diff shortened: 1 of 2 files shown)


composables/useScenarioOptions.ts
```diff
@@ -50,7 +50,8 @@ export default (parameterAxis: MaybeRefOrGetter<Parameter | undefined>) => {
       return [];
     }
     if (axis.value?.parameterType === TypeOfParameter.Numeric) {
-      return predefinedNumericOptions.value?.filter(o => o.id !== baselineOption.value?.id) || [];
+      const opts = predefinedNumericOptions.value?.filter(o => o.id !== baselineOption.value?.id) || [];
+      return opts.filter((opt, index, self) => index === self.findIndex(o => o.id === opt.id)); // dedupe by id
```
M-Kusumgar: this is completely fine for the number of options you're expected to have so happy to approve this but just an optional nitpick is that this is nicer looking code that works way worse:

1. you filter on ids
2. then you filter again so iterate through all items again
3. the filter condition is more subtle, because `findIndex` returns first index, you're filtering on first match index and youre iterating again to find that first match through the array. also because `findIndex` works from the start, you're almost always going to iterate to the element youre currently at unless its a duplicate

VS

simpler code for this (something like this, writing on github so idk if it runs)
```js
const filtered = [];
predefinedNumericOptions.value?.forEach(op => {
  if (op.id !== baselineOption.value?.id) return;
  if (!filtered.find(x => x.id === op.id)) filtered.push(op);
});
```
means you only construct one new array instead of 2 and also you only find on a deduped smaller array

M-Kusumgar (approved):
LGTM

PR #195, opened 2026-05-26
Update hospital capacity plot lines logic

david-mears-2:
# Background info

A 'capacity' is, for example, the level of hospital surge capacity considered in some scenario. (Hospital surge capacity is de facto the only kind of capacity that currently exists for daedalus, but a second theoretical example could be ICU capacity.) It may be rendered in a time series, if relevant and desired by the user, as a plot line. A 'plot line' refers to a _constant_ value as rendered in a straight line (in our case a dotted line), as opposed to the actual time series itself which is not straight:

image

# Changes in this PR

Two related updates for time series' plot lines in these two commits:

1) [Show hospital capacity plotline by default in comparison](https://github.com/jameel-institute/daedalus-web-app/commit/db08d0e66c0b83cf2e73b8b365a2a6fec282b08f):

On the comparison results page (in the time series tab), we should show the hospital capacity plot line by default, rather than hide it by default. Here is the toggle you're looking for: 
image

2) [Update hospital capacity plot lines logic: show all scenarios' capacities if comparing by capacity](https://github.com/jameel-institute/daedalus-web-app/commit/0a7f822cc918cd40c500bf9f69d69484f0477560)

One of the possible parameters by which users can perform a comparison is hospital surge capacity. When that is the axis of comparison, scenarios' capacities are going to vary, and we have the possibility of plotting multiple plot lines on a single chart. Previously, in order to avoid plotting multiple plot lines at once, the set-up was that users could determine which scenarios' plot line was rendered in the chart by hovering over that particular scenario's time series. However, in practice, this is very fiddly to pull off, using normal human levels of mouse control, especially when the the time series you're trying to hover over are very close or actually overlap (as when they have the same values). So now, the new behaviour is just to display all scenarios' capacities at once (without text labels since these would overlap and look messy). If that is too much information, the user can still use the toggle to hide the plot lines.

In the below example, each of the three compared scenarios has the same number of hospitalised individuals for every time step throughout the course of the pandemic, and so we only see a single time series in the plot (the other lines are there too, just hidden behind the pink scenario). But we still get to see where the hospital capacity plot line is for all 3.

image

### DIFF
--- a/components/Charts/Compare/CompareTimeSeries.client.vue
+++ b/components/Charts/Compare/CompareTimeSeries.client.vue
@@ -101,9 +101,15 @@ const hoveredScenario = computed(() => appStore.currentComparison.scenarios.find
   return runId === props.synchPoint?.series?.options?.custom?.scenarioId;
 }));
 
-const capacitiesForegroundedScenario = computed(() => {
-  const hoveredChartShowsCapacities = props.synchPoint?.series?.options?.custom?.showCapacities === true;
-  return hoveredChartShowsCapacities ? hoveredScenario.value : appStore.baselineScenario;
+// If the parameter by which we are comparing scenarios is a kind of capacity, then scenarios' capacities will vary, so
+// we should show capacities for each scenario. Otherwise, we only need to show capacities for the baseline scenario.
+const scenariosToShowCapacities = computed(() => {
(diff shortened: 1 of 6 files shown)


M-Kusumgar (approved):
sorry for the late review! i forgot about this one, looks good to me!

